PR #26, opened 2017-10-12
Experimental fromTemplateLiteral API.

koto:


koto:
Deprecated in v2 API.

PR #44, opened 2018-06-01
Consolidate common code in `class Trusted*` and create/call policy functions a bit more defensively.

mikesamuel:
This is not ready for commit -- it adds some tests as TODOs which
currently fail.

This is a first step towards making trustedtypes a bit more defensive.


src/enforcer.js
```diff
@@ -358,27 +408,31 @@ export class TrustedTypesEnforcer {
    * @private
    */
   getKey_(object, name) {
+    // TODO(msamuel): Can we use Object.prototype.toString.call(object)
```
koto: Is it worth it? This is to part of the private install-time API to Proxy the original setters. `object` here is a prototype of various `Element`s. If those are spoofed at the moment of running this code, we're Proxying fake DOM anyway.

src/trustedtypes.js
```diff
@@ -45,31 +50,52 @@ export const TrustedTypes = (function() {
...
+  const policyNames = selfContained([]);
 
   /**
    * Map of all exposed policies, keyed by policy name.
    * @type {Map<string,Object>}
    */
-  const exposedPolicies = new Map();
+  const exposedPolicies = selfContained(new Map());
+
+  // TODO(msamuel): common base class for Trusted*?
```
koto: Go for it, not necessarily in this PR (or here, it doesn't really matter).

src/trustedtypes.js
```diff
@@ -100,14 +127,20 @@ export const TrustedTypes = (function() {
...
-      return 'TrustedURL';
+    valueOf() {
+      return privates(this).value;
     }
   }
+  /**
+   * Name property.
+   * Required by the enforcer to work with both the polyfilled and native type.
+   */
+  defineProperty(
```
koto: Good catch! I thought `name` is just for error messages, but we actually rely on it in the polyfill enforcer.

src/trustedtypes.js
```diff
@@ -208,6 +255,12 @@ export const TrustedTypes = (function() {
       if (!policy) {
         throw new Error('Policy not found');
       }
+      // TODO(msamuel): can we require that functionName
+      // is an own property name in policy?
```
koto: It should not be required, we only call `buildType..` with constants at initialization of the API, `functionName` is not attacker-controlled.

src/trustedtypes.js
```diff
@@ -271,6 +327,10 @@ export const TrustedTypes = (function() {
    * @return {!Array<string>}
    */
   function getPolicyNames() {
+    // TODO(msamuel): Should we sort policyNames to avoid leaking or
+    // encouraging dependency on the order in which policy names are
```
koto: No opinion here. If we sort, let's do it on append, not on retrieval.

src/trustedtypes.js
```diff
@@ -289,16 +349,35 @@ export const TrustedTypes = (function() {
    * @todo Figure out if the return value (and the builder) can be typed.
    */
   function createPolicy(name, builder) {
+    // TODO(msamuel): could we also allow symbol names for non-exposed policies?
```
koto: Not really, for now we're sticking with strings, as there's a policy name whitelist in CSP, and that one relies on strings.

src/trustedtypes.js
```diff
@@ -289,16 +349,35 @@ export const TrustedTypes = (function() {
...
+    // That would simplify a lot of collision avoidance stuff in client code.
+    // For example,
+    // function toPolicyName(name) {
+    //   if ('symbol' === typeof name) {
+    //     if (getPrototypeOf(name) !== Symbol.prototype) {
+    //       throw new Error('Symbols from other Realms cannot be policy names')
```
koto: Can you copy Symbols across realms? If so, that would bypass this check.

src/trustedtypes.js
```diff
@@ -289,16 +349,35 @@ export const TrustedTypes = (function() {
...
     if (policyNames.includes(pName)) {
       throw new Error('Policy ' + pName + ' exists');
     }
+    // Register the name early so that if builder unwisely calls
+    // across protection domains to code that reenters this function,
+    // builder's author still has rights to the name.
+    policyNames.push(pName);
 
-    let innerPolicy = Object.assign({}, initialBuilder);
+    // TODO(msamuel): can the innerPolicy make do with a null prototype?
```
koto: Yes, I think we can. If this turns out to be a problem with actual real-life policies, we can relax this.

src/trustedtypes.js
```diff
@@ -289,16 +349,35 @@ export const TrustedTypes = (function() {
...
+    // That might make leaks via thisValue less likely.
+    const innerPolicy = assign({}, initialBuilder);
     builder(innerPolicy);
-
-    policyNames.push(pName);
+    // TODO(msamuel): can we freeze the inner policy post build, so that
+    // it doesn't matter as much if it leaks via thisValue?
```
koto: We can.

src/trustedtypes.js
```diff
@@ -289,16 +349,35 @@ export const TrustedTypes = (function() {
...
+    // TODO(msamuel): if so, should we await in case builder is async?
```
koto: For now, the policies expose only sync functions (because the whole DOM API is sync), but I was thinking to add `asyncCreateHTML` and such later. It will definitely have to be a separate functions in the policy, or even `document.policies.createAsync`.

tests/enforcer_test.js
```diff
@@ -189,6 +265,10 @@ describe('TrustedTypesEnforcer', function() {
 
       expect(el.src).toEqual('');
     });
+
+    it('on DOMParser.parseFromString', function() {
+      // TODO(msamuel): do we care about explicit invocations of DOMParser?
```
koto: The correct answer is ¯\_(ツ)_/¯ (issue #50)

tests/enforcer_test.js
```diff
@@ -180,6 +219,43 @@ describe('TrustedTypesEnforcer', function() {
...
+      let el = document.createElement('iframe');
+
+      expect(function() {
+        el.setAttributeNS('http://www.w3.org/1999/xhtml', 'src', TEST_URL);
+      }).toThrow();
+
+      expect(el.getAttributeNS('http://www.w3.org/1999/xhtml', 'src')).toEqual(null);
+    });
+
+    it('on copy attribute crossing types', function() {
```
koto: This test is failing, maybe it's ok to comment it out with TODO fix #47? IIUC we don't have a good idea how to address this yet, but it's not worth blocking the whole PR for that.

koto:
Hey Mike, 

I'm OK with merging this, provided the existing tests pass, LMK when you're ready with the PR.

mikesamuel:
Sorry for the delay.  Cleaning it up now.

src/enforcer.js
mikesamuel: IIUC, they are also used at restore time which happens after untrusted inputs reach the system.

I don't think it's a huge issue.
I mention it because it provides a reliable path to facts like
```js
Object.prototype.toString.call(Boolean.prototype) === '[object Boolean]'
Object.prototype.toString.call(Function.prototype) === '[object Function]'
Object.prototype.toString.call(Number.prototype) === '[object Number]'
// etc.
```

but not much moreso than typeof except for Array/Object distinctions and a few intrinsics like `Date`.

src/trustedtypes.js
mikesamuel:
```js
> class C {}
< undefined
> Object.getOwnPropertyDescriptor(C, 'name')
< {value: "C", writable: false, enumerable: false, configurable: true}
```

so the type already has a static name property that is frozen.  We just need to make it unconfigurable.

Fixed.

src/trustedtypes.js
mikesamuel: Done here.

src/trustedtypes.js
mikesamuel: Done

src/trustedtypes.js
mikesamuel: Done

src/trustedtypes.js
mikesamuel: Removed TODO

tests/enforcer_test.js
mikesamuel: Removed comment since it's tracked.

src/trustedtypes.js
koto: The compiler will usually rename the class.

src/enforcer.js
koto: I see. We can add it then (though install/uninstall are not intended to be called at all, those methods are used in tests only, to be able to reuse the document object). Though, to be fair, I should have just created a new Document object in `beforeEach` or documented the intention of those methods.

mikesamuel: I can leave this for now, or file a tracking issue.

src/trustedtypes.js
mikesamuel: A functionName need not be attacker controlled if it's not present on every instance and there's a prototype.  I think we're probably safe, but hasOwnProperty is cheap and this is policy application code, so I added it just to be safe.  Happy to rollback if that's a problem.

src/trustedtypes.js
mikesamuel: Removed

src/trustedtypes.js
mikesamuel: No.  Symbols are per-realm.

PR #108, opened 2019-01-16
Migrates spec from Google doc into bikeshed format

mikesamuel:
This migrates https://docs.google.com/document/d/1wpH37zSrfecx65TS6Up7l_WWydG_l2e6UXaZadIxzas/edit#heading=h.i8ymd0x6azcr including side notes to [bikeshed](https://github.com/tabatkins/bikeshed) format.

A follow on PR will add instructions to CONTRIBUTING and to a comment at the top of the BS file explaining how to regenerate the HTML, and will link the HTML from the project README.md.


spec/index.bs
```diff
@@ -8,16 +8,1400 @@ URL: https://wicg.github.io/trusted-types/dist/spec/index.html
...
+        return *input* and abort these steps.
+    1.  Otherwise, return stringified *input* and abort these steps.
+1.  If *input* has type *expectedType*, return stringified
+    *input* and abort these steps.
+1.  If *passThroughFunctions* is true and *input* is a {{Function}},
+    return input and abort these steps.
+1.  Otherwise, execute the following steps:
+    1.  Let *defaultPolicy* be the result of executing
+        [$Get default policy algorithm$] on *document*’s
+        [[HTML5#concept-document-bc|browsing context]]'s {{Window}}
```
vrana: What did you mean by `#concept-document-bc`? It's not in the HTML5 spec or in the Google Docs doc.

koto: Should this be https://html.spec.whatwg.org/multipage/browsers.html#browsing-context, @mikesamuel  ?

mikesamuel: No.  That define's the type browsing context, but not the algorithm for finding the one associated with the current document.

https://html.spec.whatwg.org/multipage/browsers.html#concept-document-bc is what I was trying to link to.

> A Document's browsing context is the browsing context whose session history contains the Document, if any such browsing context exists and has not been discarded, and null otherwise.

https://html.spec.whatwg.org/multipage/window-object.html#window-bc might work too.

> The Window object's browsing context is the Window object's associated Document's browsing context. It is either null or a browsing context.

PR #114, opened 2019-01-17
Remove Enforcement in timer functions from spec

vrana:
It calls `HostEnsureCanCompileStrings` which is covered below.

### DIFF
--- a/dist/spec/index.html
+++ b/dist/spec/index.html
@@ -1460,7 +1460,7 @@ c-[il] { color: #000000 } /* Literal.Number.Integer.Long */
   <div class="head">
    <p data-fill-with="logo"></p>
    <h1 class="p-name no-ref" id="title">Trusted Types Spec WIP</h1>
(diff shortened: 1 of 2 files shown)


koto:
That one is interesting. We cannot for now base the spec on `HostEnsureCanCompileStrings`, as that algorithm doesn't accept the object (usually; a string with the code) to be compiled - [TC39#938](https://github.com/tc39/ecma262/issues/938), so we can't at that point make any decision whether to reject or accept.

We did solve it in `setTimeout` / `setInterval` spec-wise, as those are web platform features. Reusing `HostEnsureCanCompileStrings` does sound like a better place to hook the spec on, but it requires changes to EcmaScript. For now I decided to leave it as-is, given that the implementation matches (`set(Timeout|Interval)` are covered; `eval` is ignored).

vrana:
I see. TT spec in [String compilation](https://wicg.github.io/trusted-types/dist/spec/#string-compilation) writes `HostEnsureCanCompileStrings` but links [EnsureCSPDoesNotBlockStringCompilation](https://www.w3.org/TR/CSP3/#can-compile-strings) which confused me. I'll send another pull-request to fix it.

vrana:
#118

PR #118, opened 2019-01-18
Fix link text in String compilation

vrana:


koto:
Actually, we need to link to either W3C [HTML5#section-hostensurecancompilestrings](https://www.w3.org/TR/html/webappapis.html#section-hostensurecancompilestrings) or WHATWG [HTML##hostensurecancompilestrings(callerrealm,-calleerealm)](https://html.spec.whatwg.org/multipage/webappapis.html#hostensurecancompilestrings(callerrealm,-calleerealm)). It's an intermediate algorithm, that now only calls the CSP one, but should first consult the types. Let's link to WHATWG, this is what we do so far in other places in the spec.

As the note says, this algorithm needs to also accept a string that ES would pass in order for this all to make sense.

koto:
You can drop the PR, I fixed it locally and will soon upload.

PR #119, opened 2019-01-18
Editorial changes in the spec

koto:
Resolving existing issues, and adding some more.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -519,7 +519,7 @@ and exposed flag (*exposed*), run these steps:
 1.  Set *policyOptions* {{TrustedTypePolicy/createURL()|createURL}}
     property to *option*'s
     {{TrustedTypePolicyOptions/createURL|createURL}} property value.
-1.  Set *policy*'s `[[options]]` internal slot value to *policyOptions*.
+1.  Set *policy*'s `[[options]]`'s internal slot value to *policyOptions*.
 1.  Set the velue for the *policyName* entry of *factory*'s `[[PolicyMap]]` to *policy*.
 1.  Return *policy*.
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1294,11 +1294,7 @@ At the beginning of step 5, insert the following steps:
 Note: See <a href="https://github.com/tc39/ecma262/issues/938">TC39/ecma262 issue #938</a>
 (adding a string to be compiled to algorithm parameters)
 
-Issue: Should we be modifying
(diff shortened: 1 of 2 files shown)


spec/index.bs
```diff
@@ -1334,14 +1325,12 @@ For the {{Element/insertAdjacentHTML()}} method, execute the
 [$Enforce a Trusted Type algorithm$] with input argument set to the *text*
 parameter value.
 
-Issue: do we need to specify that type argument should be {{TrustedHTML}}
-twice herein.
-
```
koto: Not really, this is implied from the sink definition (it accepts `HTMLString`, which is then converted in the algorithm to expectedType `TrustedHTML`)

koto:
@mikesamuel can you review?

spec/index.bs
```diff
@@ -1294,11 +1294,7 @@ At the beginning of step 5, insert the following steps:
 Note: See <a href="https://github.com/tc39/ecma262/issues/938">TC39/ecma262 issue #938</a>
 (adding a string to be compiled to algorithm parameters)
 
-Issue: Should we be modifying
-[[HTML5#hostensurecancompilestrings(callerrealm,-calleerealm)|HostEnsureCanCompileStrings]]
-or EnsureCSPDoesNotBlockStringCompilation?
-
-Modify [[CSP#can-compile-strings|HostEnsureCanCompileStrings]]
+Modify [[HTML5/webappapis#section-hostensurecancompilestrings|HostEnsureCanCompileStrings]]
```
mikesamuel: IIRC, this HTML abstract-op just calls through to the CSP one, so it seems less likely to cause problems if the change is made deeper.

spec/index.bs
```diff
@@ -530,12 +530,10 @@ and exposed flag (*exposed*), run these steps:
     with the key *policyName*, return null.
 1.  Otherwise, let *policy* be the value of the *policyName* entry of
     *factory*'s `[[PolicyMap]]`.
-1.  If *policy*'s {{TrustedTypePolicyOptions/exposed|exposed}} attribute is false,
+1.  If *policy*'s `[[options]]` {{TrustedTypePolicyOptions/exposed|exposed}} attribute is false,
```
mikesamuel: 's after options?

spec/index.bs
koto: The issue is that I want to have less integration points (to later sync the amendment to other specs), and I already integrate with HTML, so let's keep it like this.

PR #135, opened 2019-03-05
TrustedTypes.isXXX(object?) => isXXX(any)

### LINKED ISSUE #134: TrustedTypes.isHTML('test') throws exception
Native implementation in Chrome Version 74.0.3718.1 (Official Build) canary-dcheck (32-bit)
TrustedTypes.isHTML('test') throws exception if value is string

> Uncaught TypeError: Failed to execute 'isHTML' on 'TrustedTypePolicyFactory': parameter 1 ('checkedObject') is not an object.

polyfill returns false

otherdaniel:
Please have a look.

Change TrustedTypes.isXXX to accept any (rather than object?). This should resolve #134.


koto:
Can you run `npm install` and `gulp spec` to regenerate the spec HTML file? This is part of the pre-commit hook, not sure why this didn't run for you.

otherdaniel:
I couldn't push my change to the WICG repository (because of some premissions issue, I guess), so I cloned the repo, pushed the change there, and pull-requested from it. I guess my clone repo doesn't have the commit hook and hence didn't run anything.

dBat23:
Why are you making that change in the first place? Better to be specific with an object than to use (any). Maybe it didn't run because you're opening yourself up for too many bugs and duds as objects @otherdaniel

koto:
See related bug on why this is needed. For example, string is not an
object, and Trusted Types replace strings, so we expect to be called with a
string often.

dBat23:
What strings are being pulled? Parceling of an HTML?

Issue #139, opened 2019-03-15
Handle srcset attributes


koto:
Defined here: https://html.spec.whatwg.org/multipage/embedded-content.html, right now for `img` and `source` elements, but there are also plans to add it to `script` (https://github.com/whatwg/html/issues/4432) - as such, this can potentially spawn into two new types. cc @otherdaniel

mikesamuel:
One way to spec this might be to derive a spec abstraction from inputs like `[ [ trustedURL0, metadata0 ], ... ]` and a stringifier for the same to a string value that matches the syntax at [HTML5/srcset-attributes](https://html.spec.whatwg.org/multipage/images.html#srcset-attributes).

I can probably parameterize the type to make the stringifier compatible with the [`<script srcset>` proposal](https://github.com/whatwg/html/issues/4432) which would take *TrustedScriptURL*s.

1. Define a mapping from an rvalue that matches typescript type:
    ```ts
    type StructuredSrcsetInput<T extends TrustedType> = [ T, string? ][]
    ```
1. Define an IDL `interface StructuredSrcset` that mirrors that type.
1. Define an algorithm to map *StructuredSrcsetInput*s to *StructedSrcset* that takes a type
    1. Reject empty arrays
    1. Reject arrays with holes
    1. Require the URL be the right type (*TrustedURL*  vs *TrustedScriptURL*)
    1. Require the URL content not start or end with comma or HTML spaces to avoid stringifier ambiguity
    1. Stringifies the metadata and requires that it matches HTML valid floating point number with an optional letter at the end.
1. Define a *stringifier* for `StructedSrcset`
    1. Let out be an empty buffer
    1. Let isfirst = false
    1. For each element in structuredSrcset:
         1. Let url = Get(element, 0)  (TODO: Or some internal slot verbiage)
         1. Let metadata = Get(element, 1)   (TODO: Or some internal slot verbiage)
         1. If isfirst is true then
             1. Append `" , "` to empty buffer.  Note: the leading space is needed in case a URL without metadata appears in a position other than the end.
             1. Assign isfirst = false
         1. Append the URL content to out
         1. If metadata is not null then
             1. Append a space character (U+20)  to out
             1. Append metadata to out
    1. let srcsetString = ToString(out)
    1. return srcsetString
1. Spec a sink guard that chains the above together.

koto:
So, from the author's perspective (please correct me if I misunderstood the patch):
[...]

PR #140, opened 2019-03-26
enforcement for `<img srcset>`

mikesamuel:
This provides machinery relevant to the proposed solution to
issue #139.

This is not yet wired into the enforcer.

The parser is not security sensitive, but the unparser is.


dist/spec/index.html
```diff
@@ -1212,9 +1212,9 @@
 		}
 	}
 </style>
-  <meta content="Bikeshed version 9c6ae748322940994300e90a77c34437364b40a6" name="generator">
+  <meta content="Bikeshed version 86d51a2d9c3c0de4540c77189312f7e2393844af" name="generator">
```
mikesamuel: Sorry, this is cruft from running my local install of bikeshed.

mikesamuel:
Handling srcset this way adds a considerable amount of machinery.

mikesamuel:
[Drive by fix](https://github.com/mikesamuel/trusted-types/blob/3808747365aefbce2efee2aececc2737b608abee/src/enforcer.js#L682-L686) for issue #141

koto:
This is deprecated (see #139)

PR #142, opened 2019-03-27
spec: cleanup uses of `[TreatNullAs]` annotations

mikesamuel:
@tobie [notes][1]:

> As per the relevant section in the spec:
>
> > The [TreatNullAs] extended attribute must not be associated with a type
> > that is not DOMString.
>
> What you want instead is to use annotated types like so:
>
>     attribute ([TreatNullAs=EmptyString] DOMString or SomeObjectType) thing;

This reworks all uses of `[TreatNullAs]` in spec.bs to attach to `DOMString`
and preserves the note about "null"/"" confusion causing problems in the
polyfill.

[1]: https://github.com/heycam/webidl/issues/441#issuecomment-330479969


mikesamuel:
This is meant to address issue #2 but I don't know whether it'll actually solve it.

@otherdaniel Do you have any insight on whether this is worth the extra type variant?

mikesamuel:
@koto @otherdaniel ping

mikesamuel:
@otherdaniel Am I looking in the wrong place?

[Source/core/dom/Element.idl#66](https://chromium.googlesource.com/chromium/blink/+/master/Source/core/dom/Element.idl#66)
```webidl
 [TreatNullAs=NullString, CustomElementCallbacks, RaisesException=Setter] attribute DOMString innerHTML;
 [TreatNullAs=NullString, CustomElementCallbacks, RaisesException=Setter] attribute DOMString outerHTML;
```

otherdaniel:
@mikesamuel I'm afraid so. I'm not quite sure which repo that is, but the most recent change in https://chromium.googlesource.com/chromium/blink/ was ~3.5 years ago. I like to think we're more industrious than that.

This is the place where I looked:
https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/dom/element.idl?q=Element.idl&sq=package:chromium&dr&l=88-89

otherdaniel:
Ah... originally, Chromium + WebKit repos were separate, because they were very separate projects. With Blink, the structure was kept for a while, but eventually the Chromium + Blink repos were merged. (crbug.com/431458) I think you found the pre-merge Blink repo, which is no longer in use.

mikesamuel:
@otherdaniel Thanks for explaining.

I think when I converted koto's google doc to bikeshed, I went to spec docs to double check but I filed issue #154.
I'll double-check WebIDL excerpts against the spec documents and Chromium source and note any merge difficulties.

mikesamuel:
@otherdaniel PR #161 obviates this.  I did something dumb and merging recent changes into the branch is a PITA.

Issue #43, opened 2018-05-31
Consider metadata API; building blocks for HTML sanitizers


xtofian:
Browser-side HTML sanitizers are typically implemented by
 - parsing HTML into an inert `Document(Fragment)`
 - transforming the resulting "untrusted" tree into a safe tree, by pruning or copying based on attribute/element white/black-lists.
 - re-serializing the tree into HTML (not necessary if the "clean" tree can be directly attached to the DOM).

Examples: https://google.github.io/closure-library/api/goog.html.sanitizer.HtmlSanitizer.html, https://github.com/cure53/DOMPurify.

The necessary whitelists are a fairly significant part of the sanitizer's code size (blacklists would be smaller, but blacklists are generally brittle); https://github.com/cure53/DOMPurify/blob/master/src/attrs.js, https://github.com/cure53/DOMPurify/blob/master/src/tags.js, https://github.com/google/closure-library/blob/master/closure/goog/html/sanitizer/attributewhitelist.js, https://github.com/google/closure-library/blob/master/closure/goog/html/sanitizer/tagwhitelist.js

TT introduces implicit knowledge into the DOM API of the security semantics of DOM attributes. It seems worth considering if we can expose this information somehow, so that it can be used to as the basis of a HTML sanitizer's policy, which could then avoid the code size overhead of their own metadata tables.

Caveat:  This will likely not obviate the need for sanitizer-specific whitelists/blacklists altogether: There are elements whose attributes don't require sanitization in a typical application's threat model (where the presence of the element in the DOM is application controlled, and only the attribute's value is potentially malicious); while the entire element is undesirable to keep when sanitizing entirely-untrusted HTML markup.  E.g., `<form>` comes to mind.

koto:
This is discussed under https://github.com/WICG/purification/issues/1.

koto:
Apart from minimizing the code of the sanitizers, the metadata might be helpful for other libraries. For example, Angular compiles the template (in [AOT mode](https://angular.io/guide/aot-compiler) at build time) that later on are serialized to a series of definitions based on strings. Later on, these strings are serialized into DOM using wrappers over `createElement` and `setAttribute` calls here:

https://github.com/angular/angular/blob/master/packages/core/src/view/element.ts#L151

[...]

PR #149, opened 2019-04-30
Added a first draft of the type metadata.

koto:
Addresses issue #43.


dist/cjs/trustedtypes.api_only.js
```diff
@@ -278,6 +434,91 @@ const trustedTypesBuilderTestOnly = function() {
...
+    // TODO: Support namespaces.
+    const canonicalAttr = toLowerCase.apply(String(attribute));
+    return getTypeInternal_(tagName, 'attributes', canonicalAttr);
+  }
+
+  /**
+   * Returns a type name from a type map.
+   * @param {string} tag A tag name.
+   * @param {string} container 'attributes' or 'properties'
+   * @param {string} property The attribute / property name.
```
mikesamuel: s/property/name/ ?

dist/cjs/trustedtypes.api_only.js
```diff
@@ -278,6 +434,91 @@ const trustedTypesBuilderTestOnly = function() {
...
+   * @return {string|undefined}
+   * @private
+   */
+  function getTypeInternal_(tag, container, property) {
+    if (HTML_TYPE_MAP['*'][container][property]) {
+      return HTML_TYPE_MAP['*'][container][property];
+    }
```
mikesamuel: I see that '*' has little on it so this seems ok as written.

But here, the general rule trumps a specific rule which is the reverse of what we've done for things like https://github.com/Polymer/polymer-resin/blob/master/lib/contracts/contracts.js which has a general rule for `*[src]` -> TrustedScriptURL and specific ones for `img[src], video[src], audio[src]` -> TrustedURL.

Is there a maintenance hazard due to that?

dist/cjs/trustedtypes.api_only.js
```diff
@@ -278,6 +434,91 @@ const trustedTypesBuilderTestOnly = function() {
...
+
+    const canonicalTag = toUpperCase.apply(String(tag));
+
+    if (!HTML_TYPE_MAP[canonicalTag]) {
```
mikesamuel: Ought this be a hasOwnProperty check?

Or ought the copy pass above do Object.setPrototypeOf(x, null) for all objects where properties correspond to tags or property/attribute names?

Or just use Map instead of {}.

This also affects the [property] lookups above and below.

dist/cjs/trustedtypes.api_only.js
```diff
@@ -278,6 +434,91 @@ const trustedTypesBuilderTestOnly = function() {
...
+   * @param {string} tagName The name of the tag of the element.
+   * @param {string} attribute The name of the attribute.
+   * @param {string=} elementNs Element namespace.
+   * @param {string=} attributeNs The attribute namespace.
+   * @return {string|undefined} Required type name or undefined, if a Trusted
+   *   Type is not required.
+   */
+  function getAttributeType(tagName, attribute, elementNs = '',
+      attributeNs = '') {
+    // TODO: Support namespaces.
```
mikesamuel: It looks like namespace support should probably go into getTypeInternal_.

Ok, so for `<svg:a xlink:href>` you have two distinct namespaces in play.
Are namespaces URLs, or common prefixes?

I wonder whether it'd be easier to code-generate a flat table that maps strings like '*/*/*/attr' would mean any element namespace, any element, any attribute namespace, attribute named attr.

Then, for given values we could cascade to different properties with fewer object, own property checks.  So for `<svg:a xlink:href>` we might try some of (I'm not sure exactly which) in order

1. `"svg/a/xlink/href"`
1. `"*/a/xlink/href"`
1. `"*/*/xlink/href"`
1. `"svg/a/*/href"`
1. `"*/a/*/href"`
1. `"*/*/*/href"`

and avoid a bunch of code that pulls out objects, tests whether they're not null and keeps looking inside.

src/enforcer.js
```diff
@@ -64,71 +64,40 @@ function isHttpUrl_(url) {
...
+/**
+ * Converts an uppercase tag name to an element constructor function name.
+ * @param {string} tagName
+ * @return {string}
+ */
+function convertTagToConstructor(tagName) {
+  if (tagName == '*') {
+    return 'HTMLElement';
+  }
+  if (['AUDIO', 'VIDEO'].includes(tagName)) {
```
mikesamuel: It seems like some Object.getPrototypeOf should let you determine where setters are defined.

dist/cjs/trustedtypes.api_only.js
koto: I guess we can revert this in code (first look in the element, then in '*'). For current values this doesn't seem to cause any  behavior change. FWIW `*[src]` in not a Trusted Type at all, we only cover `src` for selected elements (e.g. because of Web Components that may create their own attributes with arbitrary semantics).

src/enforcer.js
koto: Good idea, will fix.

dist/cjs/trustedtypes.api_only.js
koto: We probably can't use prefixes, this should be based on namespace URIs, which makes it a bit cumbersome to work with. Either the lookup key is not a single string, or we use a separator that cannot be present in a URL, like ` ` . 
The goal here is to end up with the API shape that's easiest to work with when creating DOM (and then you either have a tag name, attribute name, or the respective nodes already). Thankfully [`namespaceURI`](https://developer.mozilla.org/en-US/docs/Web/API/Node/namespaceURI) for both attribute and element nodes returns the canonical URL.

FWIW in current Chrome/Firefox in non-pure-XML documents it seems that changing the prefix of a namespace via `xmlns` makes them lose their meaning (check http://html5sec.org/test/#87 and try to rename xlink), which is reflected by `namespaceURI` returning null. It seems like some of the prefixes are simply hardcoded for X/HTML documents, that's probably specced somewhere. Nevertheless, in pure `text/html` documents namespaces can be effectively renamed, so we cannot rely on a prefix.

As a sidenote, we might be able to just avoid the problem altogether by making sure xlink just stops being risky, period. It's already deprecated: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xlink:href

src/enforcer.js
koto: Actually, abandoned the idea - it would increase the complexity of the function installing the set wrapper (with some tiny runtime cost too), and we fully control the element list anyway, so this should not be a large maintenance burden to keep this in sync with DOM API (which we have to anyway). Keeping this as is, will refactor if this turns out to be an issue.

dist/cjs/trustedtypes.api_only.js
koto: I implemented the namespaced attribute enforcement. PTAL.

src/enforcer.js
koto: This is now implemented as you suggested (the logic is in `wrapSetter_` for properties, and for attributes `setAttribute` checks with the API (`TT.getAttributeType`) based on a tag name.

PR #157, opened 2019-05-07
Added BrowserStack support (runs on Travis). 

koto:


dist/spec/index.html
```diff
@@ -1,7 +1,7 @@
 <!doctype html><html lang="en">
  <head>
```
manningj917-eng: Joseph Manning 9933 e. railroad av. Terre HAUTE in.47803 d-o-b  12-13-1970     8125814195

PR #158, opened 2019-05-07
Adds use case to spec section 1.3.

mikesamuel:
Addresses

> Issue 3: Something about reducing the review surface

Some more motivation might be timely since this is getting more eyes.


spec/index.bs
```diff
@@ -82,9 +82,39 @@ Issue: figure out if we need this section here. CSP has none, SRI has one and it
...
+        run with enforcement enabled.
+
+    The team consolidates the policy implementations and the
+    safe abstractions that use them in a small number of heavily
+    reviewed files and requires extra approval for commits that
+    affect these files.
+
+    The need to create trusted values to affect [=injection sinks=],
+    combined with additional scrutiny on changes that affect policy
+    code, incents developers to find novel ways to use safe
```
koto: incents developers to use safe abstractions (consider removing "to find novel ways", that to me sounds like finding creative abuse methods, or bypasses).

spec/index.bs
```diff
@@ -82,9 +82,39 @@ Issue: figure out if we need this section here. CSP has none, SRI has one and it
     libraries that perform auxiliary tasks (e.g. analytics, performance
     monitoring). To ensure that none of these components introduces DOM
     XSS vulnerabilities, author defines a Trusted Type policy in the
-    templating policy and enables the enforcement for the DOM sinks.
-*   Issue: Something about reducing the review surface
-
+    templating policy and enables enforcement for the DOM sinks.
+*   A large team maintains a complex client-side application.
+    They create a number of Trusted Types policies so that
```
koto: I think you could remove the details of how the policies behave, and focus that there is a few of them and that they are reviewed. The current description might be too detailed for this section I feel.

koto:
Obsoleted by #335.

Issue #65, opened 2018-06-22
Consider dropping TrustedURL


koto:
We started integrating the application with the polyfill. So far the biggest obstacle is the `TrustedURL` enforcement. It turns out linking to other content is common in the web (who knew?).

Guarding how application produces URLs is important for XSS prevention because of a presence of scriptable protocols like `javascript:` (same-origin) or `data:` (cross-origin). Historically, there were also other URL schemes like `jar:` that could result in an XSS.

It's also desirable for non-XSS related reasons. Some examples:
* navigating to custom URL handlers might initiate actions harmful to users (e.g. [`tel:`], (https://tools.ietf.org/html/rfc3966#section-11), [`facetime:`](https://medium.com/section-9-lab/abusing-ios-url-handlers-on-messages-96979e8b12f5), android intents, `chrome-extension://`)
* some applications may want to prevent directing their users to external sites
* defacement / spoofing risks when loading subresources from 3rd party sites
* IP address of the user may be disclosed, together with other fingerprintable data, when a `http:` resource is fetched from a `https:` document
 * creating `iframes` from malicious URLs may enable exploitation of other bugs (e.g. when a `postMessage` channel is established without verifying peer origins).
* loading subresources from different origins into the same renderer process might data exfiltration via Spectre-like bugs.
* attacker-controlled stylesheets may [exfiltrate data from the document]((http://sirdarckcat.blogspot.com/2008/10/about-css-attacks.html)
* pointing a form to external URL via `action` or `formaction` attributes may exfiltrate data (usually, credentials)
* [`<base>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base) URL might be used to hijack all relative links.

However, in practice, for DOM XSS prevention alone it's enough if following the URL will not execute a script in a same-origin document. The check for this is simply - after parsing, and making the URL absolute, make sure it's not a `javascript:` one. In other words, there's no user-controlled sanitization required for `a.href` and other standard URL sinks. It's enough if the document behaves like it would under `script-src *`, without `unsafe-inline` keyword. 

In that spirit, we might simply drop the `TrustedURL` type and mandate that, under TT enforcement the host environment disables `javascript` URLs (like if there was a `script-src` without `unsafe-inline`). Such behavior is polyfillable (e.g. `URL` may be used to correctly parse a URL and extract the protocol). This offers a very simple DOM XSS containment (only policies may introduce DOM XSS) setting:

`Content-Security-Policy: trusted-types [policy-list]`

[...]

PR #165, opened 2019-05-08
Removed url-allow-http policy support.

koto:
This experimental feature never made it to the spec draft, and is probably better left as a default policy.


koto:
Linked to #65.

PR #168, opened 2019-05-09
Added TrustedTypes extended attribute to spec

koto:
Right now used by the attribute setters in https://wicg.github.io/trusted-types/dist/spec/index.html#enforcement-in-sinks, but this attribute will come in handy to describe the metadata API in #149.

Once multiple values are allowed for the extended attribute, we might also solve the type<->sink allocations.


koto:
@otherdaniel Please take a look at this, I'm especially concern about the possible implementation hurdles around this (I want something that doesn't worsen the maintainability of the implementation).

Issue #144, opened 2019-04-03
adjustments to HostEnsureCanCompileStrings discard value


mikesamuel:
https://github.com/WICG/trusted-types/blob/1f0e41ff742de43e70261814ac2b2d314b630e1e/spec/index.bs#L1430-L1446

This uses *Get Trusted Type compliant string* which might apply the default policy.
This is the right thing to do so that applications can provide minimal workarounds for legacy code patterns like

```js
const globalObject = new Function('return this;')();
```

which appears frequently in JS library code.

----

Uses in the JS spec are written like

```md
1. Perform ? HostEnsureCanCompileStrings(evalRealm, evalRealm).
```

The return a value is never used: it either throws or returns normally.

----

This causes two problems:

1.  Any adjustments made by the policy are lost, so a policy like
    ```js
    TrustedTypes.createPolicy('default', {
      createScript(x) { return rewriteJavascript(x); }
    });
    ```
    the value that will reach CreateDynamicFunction will be `x`, not `rewriteJavascript(x)`.
1.  Any non-trusted types value will be stringified twice which will cause unnecessary
    visible side-effects for values like
    ```
    let stringifyCallCount = 0;
    f = new Function({
      toString() {
        return stringifyCallCount++ ? 'evil()' : 'good()'
      });
    ```
    which might cause the policy to see 'good()' but CreateDynamicFunction to receive 'bad()'.

mikesamuel:
AFAICT, there's no fix that satisfies the `const globalObject = new Function('return this;')();` use case and which does not require changing the nature of the JS / browser interactions.

@mikewest Do you have a sense of how browser implementors might react to replacing language like

> 1. Perform HostEnsureCanCompileStrings(*calleeRealm*, *callerRealm*, *goal*, *sourceText*).

with

> 1. Set *sourceText* to HostBeforeCompile(*calleeRealm*, *callerRealm*, *goal*, *sourceText*).

(https://github.com/tc39/ecma262/pull/1498 defines *goal* and *sourceText*.)

koto:
Should this be tracked under https://github.com/WICG/trusted-types/issues/207 or the tc39 repos, or do we need this bug separately?

PR #170, opened 2019-05-09
Rewrote CSP & EcmaScript integration

### LINKED ISSUE #143: Clarify interaction between unsafe-eval and TrustedScript.
@lweichselbaum @koto 

Per https://twitter.com/we1x/status/1113340867409076224 since TT with `eval`/`Function` guards provides similar protection to CSP without unsafe eval, maybe it's worth clarifying how an application might provide degraded service when TT is not available but take advantage of TT where it is.

----

Should a CSP policy without `unsafe-eval` prevent `eval(myTrustedScriptValue)`?

I'm leaning no since `unsafe-eval` in a CSP header can't be contingent on trusted-types being supported.

----

If `unsafe-eval` is present, does trusted-types guard `eval(x)`/`Function(x)`/`import(...)` at all?

----

Similarly for `wasm-unsafe-eval`?

mikesamuel:
This reworks how `new Function(source)` and `eval(source)` are
checked against a CSP and Trusted Types policy.

It trusts TrustedScript when the relevant CSP policies and TrustedType
configurations agree on whether to enforce and the TrustedType
configuration places limits on policy creation.

It also changes the previous language to use *calleeRealm* instead of
*callerRealm* for consistency with other sinks.

> ```js
>   let f = new self.top.Function(source);
> ```
> In this case, the *callerRealm*'s Window is `self` and the
> *calleeRealm*'s Window is `self.top`.
> The Trusted Types portion of this algorithm uses *calleeRealm*
> for consistency with other sinks.
> ```js
>   // Assigning a string to another Realm's DOM sink uses that
>   // Realm's default policy.
>   self.top.body.innerHTML = 'Hello, World!';
>   // Using another Realm's builtin Function constructor should
>   // analogously use that
>   // Realm's default policy.
>   new self.top.Function('alert(1)')()
> ```

It also makes recent versions of `bikeshed` run without warnings.

Fixes #143
Issue #144

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -160,7 +160,7 @@ producing types in parts of the application, while still using and
 accepting strings in other parts of the codebase). In that sense,
 Trusted Types are backwards-compatible with the regular DOM APIs.
 
-### TrustedTypes extended attribute
+### TrustedTypes extended attribute ### {#!trustedtypes-extended-attribute}
 
 To ensure that Trusted Types are required in DOM property setters [=injection sinks=], we introduce
 <dfn data-dfn-type="extended-attribute" data-x="TrustedTypes" data-lt="TrustedTypes"><code>[TrustedTypes]</code></dfn>
@@ -986,7 +986,7 @@ Given a {{Response}} (|response|), perform the following steps:
     1.  Otherwise, if |item| is an
         <a href="https://tools.ietf.org/html/rfc8259#section-4">object</a>,
         for each |name| → |value| of |item|, execute the following steps:
-        1.  Let |propertyName| be the result of running ASCII lowercase on name.
+        1.  Let |propertyName| be the result of running |=ASCII lowercase=| on |name|.
         1.  Execute the first statement, switching on |propertyName|:
             1.  case `“dom”`:
                 If value is a valid {{StringAtSinkDisposition}}, set
@@ -1322,8 +1322,7 @@ At the beginning of step 5, insert the following steps:
     [$Get Trusted Type compliant string$] algorithm, with
     *   *document* set to the owner document of *eventTarget*,
     *   *input* set to *value*,
-    *   *expectedType* set to {{TrustedScript}}, and
-    *   |passThroughFunctions| set to false.
+    *   *expectedType* set to {{TrustedScript}}
 1.  If the algorithm throws an error, abort these steps.
 
 
@@ -1398,26 +1397,17 @@ throws an "`EvalError`" if not:
     [$Get Trusted Type compliant string$] algorithm, with:
     *   |document| as |document|,
     *   |source| as |input|,
-    *   {{TrustedScript}} as |expectedType|, and
-    *   false as |passThroughFunctions|.</ins>
+    *   {{TrustedScript}} as |expectedType|.</ins>
 
 3.  <ins>If the algorithm throws an error, throw an {{EvalError}}.</ins>
 
-4.  Let |globals| be a list containing |calleeRealm|'s [=Realm/global object=] and |calleeRealm|'s
-    [=Realm/global object=].
-
-5.  <ins>Let |cspPolicies| be an empty list.</ins>
-
-6.  <ins>For each |global| in |globals|:
-
-    1.  For each |policy| in |global|'s [[CSP#global-object-csp-list|CSP list]]:
-        1.  Add |policy| to |cspPolicies|.</ins>
-
-7.  <ins>Let |isExempt| be the result of executing the [$IsSourceExempt$] algorithm with:
-    *   |cspPolicies| as |cspPolicies|, and
+4.  <ins>Let |isExempt| be the result of executing the [$IsSourceExempt$] algorithm with:
     *   |document|’s {{TrustedTypeConfiguration}} as |ttConfig|.</ins>
 
-8.  For each |global| in |globals|:
+5.  Let |globals| be a list containing |calleeRealm|'s [=Realm/global object=] and |calleeRealm|'s
+    [=Realm/global object=].
+
+6.  For each |global| in |globals|:
 
     1.  Let |result| be "`Allowed`".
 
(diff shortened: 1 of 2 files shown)


mikesamuel:
@mikewest This is what we discussed via email.

@otherdaniel @koto Fyi, this has a normative change to the realm used.

spec/index.bs
```diff
@@ -1518,20 +1500,184 @@ enum SupportedType {
...
+1.  If |ttConfig|'s {{TrustedTypeConfiguration/unknownPolicyName}} is
+    {{UnknownPolicyNameDisposition|"allow"}} then
+    set |isTTStrict| to `false`.
+1.  Otherwise if |ttConfig|'s {{TrustedTypeConfiguration/allowedNames}} contains `'*'` then
+    set |isTTStrict| to `false`.
+1.  Let |isTTEnforced| = `false`.
+1.  If |ttConfig|'s {{TrustedTypeConfiguration/domSinks}} is
+    {{StringAtSinkDisposition|"reject"}} then
+    set |isTTEnforced| to `true`.
+1.  Let |isCSPEnforced| = `false`.
```
koto: The name might be a bit misleading, IIUC this is not checking the policy disposition, but whether any of the CSPs (enforcing, or report-only) cover `script-src` without `'unsafe-eval'`.

spec/index.bs
```diff
@@ -1518,20 +1500,184 @@ enum SupportedType {
...
+        Otherwise if |policy| contains a [[CSP3#directive|directive]]
+        whose [[CSP3#directive-name|name]] is "`default-src`", then set
+        |source-list| to that [[CSP3#directive|directive]]'s
+        [[CSP3#directive-value|value]].
+    1.  If |source-list| is not `null`, and does not contain a
+        [[CSP3#source-expression|source expression]] which as an
+        <a>ASCII case-insensitive</a> match for
+        the string "<a grammar>'unsafe-eval'</a>", then
+        set |isCSPEnforced| to `true`.
+1.  If |isTTStrict| is `true` and |isCSPEnforced| is equal to
```
koto: Why do you calculate & check for `isCSPEnforced` here? Can't `eval(TrustedScript)` be exempt from CSP restrictions if `isTTStrict` and `isTTEnforced` under script-restricting CSP, and always otherwise? 

IIUC, that would happen if you don't do any CSP-related calculations in this algorithm. `EnsureCSPDoesNotBlockStringCompilation` will enforce any CSP restrictions if this algorithm returned false, which may simply do nothing if there are no CSPs set.

What usecase does the condition in this step cover?

spec/index.bs
mikesamuel: Um.  isCSPEnforced should mean

∃ p. !isReportOnly(p)  and  'unsafe-eval' not in p

If not, that's a spec bug.
What do you think of the name given that goal?

spec/index.bs
mikesamuel: I'm looking for consistency between enforcement, because I want testing with TT & CSP in report-only mode to only report violations during normal use that need to be fixed to avoid violation reports with enforcement on during normal use.

So I worry that if I don't check for consistency between TT and CSP that an engineer migrating would see CSP report-only violation reports that would be suppressed by the is-exempt check were TT and CSP both in enforce mode.

If that sounds like a good goal, maybe this deserves a note.

spec/index.bs
koto: The name is fine (there's just no check for report only here, I think).

spec/index.bs
koto: So this assumes you're swapping the CSP enforcement together with TT enforcement? For example, if you have enforcing TT and report-only CSP, this will generate violations that you have no way of fixing, which would disappear when CSP is enforced. 

In general, during rollouts it would be rare to have the disposition of both CSP and TT to be changed simultaneously.

mikesamuel:
I rebased and rewrote to recognize `script-src: 'trusted-script'`.

mikesamuel:
Ack.  Just noticed that I need to rewrite IsSourceExempt.  It should simplify a lot since there's no need to distinguish between strong and weak TT configurations.

spec/index.bs
```diff
@@ -1369,6 +1350,135 @@ interface DOMParser {
...
+
+        2.  If |policy| contains a [=directive=] whose [=directive/name=] is "`script-src`", then
+            set |source-list| to that [=directive=]'s [=directive/value=].
+
+            Otherwise if |policy| contains a [=directive=] whose [=directive/name=] is
+            "`default-src`", then set |source-list| to that directive's [=directive/value=].
+
+        3.  If |source-list| is not `null`, and does not contain a [=source expression=] which is
+            an [=ASCII case-insensitive=] match for the string "<a grammar>`'unsafe-eval'`</a>", <ins>and
+            it is not the case that (|isExempt| is true and |source-list| contains an
```
koto: Would the language be more clear if you reverted the condition?

and (isExempt is false or source-list does not contain a match for 'trusted-script')?

There's a fair amount of double negations here :)

mikesamuel: So instead of

    sourceList != null && !sourceExpression.containsCaseInsensitively("'unsafe-eval'")
    && !(isExempt && sourceList.containsCaseInsensitively("'trusted-script'"))

it would be

    sourceList != null && !sourceExpression.containsCaseInsensitively("'unsafe-eval'")
    && (!isExempt || !sourceList.containsCaseInsensitively("'trusted-script'"))

which would correspond to

> If |source-list| is not `null`, and does not contain a [=source expression=] which is
> an [=ASCII case-insensitive=] match for the string "<a grammar>`'unsafe-eval'`</a>", <ins>and
> it is the case that (|isExempt| is not true or |source-list| does not contain a
> [=source expression=] which is an [=ASCII case-insensitive=] match for the
> string "<a grammar>`'trusted-script'`</a>")</ins> then

?

koto: Yes.

mikesamuel: Done.

Issue #123, opened 2019-01-26
open for extension in createPolicy last arguments 


Jxck:
Hi

```js
TrustedTypePolicy createPolicy(DOMString policyName, TrustedTypeInnerPolicy policy, optional boolean expose = false);
```

optional boolean seems better to use object for future extension.

for example,  addEventListener has same problem when adding new options.

```js
// before
document.addEventListener('touchstart', handler, true);

// after
document.addEventListener('touchstart', handler, {capture: true});
```

if you needs feature detection, you need to do like this.

```js
var supportsPassive = false;
try {
  // define getter for opts.passive
  var opts = Object.defineProperty({}, 'passive', {
    get: function() {
      // if called, option obj for 3rd arguments ara supported
      supportsPassive = true;
    }
  });
  // checking for supports or not
  window.addEventListener("test", null, opts);
} catch (e) {}
```

for avoid this tragic history, its better to think about make it object for last argument, even if you currently don't think about another option without `expose`.

like this

```js
TrustedTypePolicy createPolicy(DOMString policyName, TrustedTypeInnerPolicy policy, optional TrustedTypeOption option);

interface TrustedTypeOption {
  boolean expose;
}
```

xtofian:
Agree with your point -- options are better off in a structured form from the start.  

For the `exposed` option specifically, perhaps we should just get rid of the extra boolean parameter of createPolicy, since `exposed` is already present as a field in the second parameter, [TrustedTypePolicyOptions](https://wicg.github.io/trusted-types/dist/spec/#trusted-type-policy-options)

koto:
In fact, we're thinking of whether the exposed argument is needed at all. FWIW only the default policy is exposed usually, and the users have the option of exposing other policies via regular JS means anyway (e.g. via assigning to a global variable). That would simplify the design a bit.

koto:
Closing for now, as we're dropping the `exposed` parameter, but once/if additional parameters will be added to the API, we'll use a dictionary of options. Thanks for pointing that out!

PR #171, opened 2019-05-14
Removed getExposedPolicy() and the exposed flag from createPolicy().

koto:
See #123.


koto:
Chrome implementation tracked in https://bugs.chromium.org/p/chromium/issues/detail?id=992006.

PR #200, opened 2019-07-24
Trimming the violating value to 40 characters, not the whole sample.

koto:


koto:
Chrome impl tracked in https://bugs.chromium.org/p/chromium/issues/detail?id=991195

Issue #176, opened 2019-05-28
Putting guards at primitives instead of sinks


annevk:
I discussed this briefly with @koto and thought I'd file it to not lose track of the idea. What if instead of sinks we add the guards at the various primitives. E.g., not `<a>`, `<form>`, etc. but "navigate". Not `fetch()`, `<img>`, etc. but "fetch". Not `appendChild()` et al but "prepare a script" (or some such). Not `innerHTML` and friends but "HTML fragment parsing algorithm".

This would put the actual protections right at the dangerous points. We'd still have to change sinks to allow for typed objects to reach the dangerous points, but there's no longer the issue of overlooking a sink or overlooking trusted types when adding a new sink. Or the issue of it not being clear how to update all the various sinks as with `Location` as we could opt not to add trusted types for all of them.

koto:
That's a really good idea! It's mostly just an exercise in finding the right primitives matching the current sinks. For example, for now `fetch()` is not guarded. I assume the default policy (or a callback function) would run not at a sink setter level, but at the lower-lever primitive (like a spec algorithm). 

Two potential downsides I see is that a) it might make the integration with other spec more verbose, so more challenging and b) maybe not everything is polyfillable. 

I'll take a look at where we could hook the current TT sink setters, and whether that causes some change of behavior (it will for individual `Location` properties, so for `<a href>` as well, but that's actually what we want - #64).

koto:
One problem with the Fetch integration is that this would happen too late. To avoid having to taint track nodes, TT are guarded at node attribute setters, even if the node is not connected to the document. 

For example, for scripts fetch would only be triggered when the script is [prepared](https://html.spec.whatwg.org/multipage/scripting.html#prepare-a-script), but we'd rather require the type earlier, in this case when setting the `src`. 

[...]

Issue #169, opened 2019-05-09
Cover missing sinks


koto:
* HTMLElement.manifest
* Link.imagesrcset
* (a|area|svg:a) ping (space separated URLs)

There's also some URLs that don't seem to be used by the user agent in any way:
* blockquote cite
* \* itemid
* \* itemprop
* \* itemtype

koto:
Blob constructor as well. We should require `TrustedHTML` if the Blob type is one of the HTML types. It seems fine to allow for contructing Blobs from non-DOMstring arguments.

While usually the blobs are used in new documents, like iframes, where they can simply trigger in-origin XSS from user input, they can also be used to create nodes from strings like this:

```
b = new  Blob([`<script>alert(1)</script>`], {type: 'text/html'})
xhr = new XMLHttpRequest();
xhr.responseType = 'document'; 
xhr.open('GET', URL.createObjectURL(b), true); 
xhr.send();
// ...
xhr.response.scripts[0] // a new script node.
```

koto:
The sinks mentioned in https://github.com/w3c/webappsec-trusted-types/issues/169#issue-442315842 self-deprecated with `TrustedURL`. 

Blob is complicated, and it seems like TT can't solve it alone. In general, TT are bypassable with a colluding same-origin document (https://w3c.github.io/webappsec-trusted-types/dist/spec/#cross-document-vectors) - a problem that `Blob`s surface. It looks like a better primitive would be to make Blobs inherit the creator's policies - https://github.com/w3c/FileAPI/issues/142, or allow for cross-origin/sandboxed blob creation, perhaps by default.

PR #204, opened 2019-08-06
Replace TrustedURL with calling a default policy on navigation to javascript: URLs.

### LINKED ISSUE #64: Bypass via HTMLAnchorElement properties
Found by @sirdarckcat:

It's still possible to execute the JS bypassing the policy by directly manipulating the `HTMLAnchorElement` properties like protocol, pathname etc.

```javascript
a.href = TrustedTypes.createPolicy('foo', (p) => {
  p.createURL = (d) => d // actually sanitize here, but that's not relevant for the bypass.
}).createURL('http://notevil.com');

a.pathname='\nalert(1)';
a.protocol='javascript:';
a.click();
```

Full setup:
```text
data:text/html,<meta http-equiv="Content-Security-Policy" value="trusted-types *"> <script src="https://wicg.github.io/trusted-types/dist/es6/trustedtypes.build.js"></script><a id=a>clickme</a><script>a.href = TrustedTypes.createPolicy(Math.random(), (p) => {p.createURL = (d) => d}).createURL('http://notevil.com');a.pathname='\nalert(1)';a.protocol='javascript:';a.click()</script>
```

Given that this can be abused by multiple properties, and each property requires only a part of the URL, it's not clear how the fix might look like, short of disabling the APIs, or just disallowing changing a.protocol (to `javascript:` or totally).

https://developer.mozilla.org/en-US/docs/Web/API/HTMLAnchorElement mentions that these properties are experimental.

koto:
This removes the burden from all authors to create types when interacting with common sinks that usually don't cause DOM XSS (unless for javascript: URLs).

This PR prevents `javascript:` URLs from working by default, and allows programmatic opt-in to enable them one-by-onefor the few applications that need them.

Related to #176. 
Partially addresses #169.
Fixes #64.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -105,8 +105,8 @@ which are substantially easier to safeguard, monitor and review.
 
 *   An existing web application interacts with the DOM mostly using XSS-safe
     patterns (i.e. withour using [=injection sinks=]). In a few places, however,
-    it resorts to using risky patterns like calling into `innerHTML`, `eval`, or
-    creating `javascript:` URIs.
+    it resorts to using risky patterns like loading additional script using
+    JSONP, calling into `innerHTML` or `eval`.
 
     Review finds that those places do not cause XSS (e.g. because
     user-controlled data is not part of the input to those sinks), but it's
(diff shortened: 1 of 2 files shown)


spec/index.bs
```diff
@@ -1327,6 +1221,37 @@ Content-Security-Policy: trusted-types one two default
...
+    * |encodedScriptSource| as |value|
+    * `"TrustedScript` as |trustedTypeName|
+    * &laquo; `"Location.href"` &raquo; as |arguments|
+1. If |convertedScriptSource| does not have a type {{TrustedScript}}, return `"Blocked"` and abort further steps.
+1. Set |urlString| to be the result of prepending `"javascript:"` to stringified |convertedScriptSource|.
+1. Let |newURL| be the result of running the [=URL parser=] on |urlString|. If the parser returns a failure, return `"Blocked"` and abort further steps.
+1. Set |request|'s [=request/url=] to |newURL|.
+
+     Note: No other CSP directives operate on `javascript:` URLs in a pre-navigation check. Other directives that check javascript: URLs
+           will operate on the modified URL later, in the [=inline check=].
```
mikewest: Do you want this to have the `eval()` behavior, wherein the policy's acceptance of the script by minting a `TrustedScript` allows it to bypass the absence of `unsafe-eval`? Or is it expected that a policy which disallows inline script will continue to disallow `javascript:` navigation?

koto: We could control this via [`trusted-script`](https://wicg.github.io/trusted-types/dist/spec/#trusted-script-csp-keyword) keyword in `script-src`. I.e. if there are scripty directives in your CSP, all should have `trusted-script`.

koto: I added the description on how that would behave with `'trusted-script'` CSP keyword. The possible complexity I see is that trusted-script depends the whole CSP lists (like it was the case for eval). I feel we can improve this later on, perhaps by requiring that 'trusted-script' is present on every CSP. This feels like better to be discussed in the CSP bugtracker later on, or in the ocntext of CSP next?

PR #206, opened 2019-08-08
Add trustedTypes.defaultPolicy getter. Fixes #185.

### LINKED ISSUE #184: Specify the violation events

### LINKED ISSUE #185: Make default policy available to tools
Tools like a template language might want to apply a policy to string inputs, akin to applying the default policy, for example to vet strings that reach `<a href={{x}}>`.  They might benefit from being able to default to the document's default policy.

Maybe expose an `[Unforgeable]` property `TrustedTypes.defaultPolicy` with type `null | TrustedTypesPolicy`.

koto:


koto:
Chrome impl in https://bugs.chromium.org/p/chromium/issues/detail?id=992031

PR #214, opened 2019-09-10
Don't throw on policy collision when policy names = *

Siegrift:


tests/trustedtypes_test.js
```diff
@@ -562,6 +552,14 @@ describe('v2 TrustedTypes', () => {
       expect(TrustedTypes.getPolicyNames()).not.toContain('default');
       expect(TrustedTypes.getPolicyNames()).toContain('a');
     });
+
+    it('allows policy collision only when allowed policies includes *', () => {
```
koto: Can you create 2 separate test cases? One with `setAllowedPolicyNames(['*'])` and another one with `setAllowedPolicyNames(['foo'])` ?

koto: "policy collision" => "policy name collision"

Siegrift: done.

Siegrift: Yeah, good idea. I split the tests and also moved them in a separate `describe`.

PR #236, opened 2019-11-08
Alternate take for script enforcement. 

koto:
Re: whatwg/dom#789 and whatwg/html#3052

As proposed by @annevk, add slots for script URL / text, populate them
when calling sink functions, and verify them when a script is prepared,
optionally running a default policy on a value read from the DOM if
it's different than the slot value.

It avoids integration points with DOM mutation algorithms, but we still
need to support `script.setAttribute('src')`, which is missing here.


spec/index.bs
```diff
@@ -1034,44 +1338,28 @@ partial interface HTMLObjectElement : HTMLElement {
...
 
-Additionally, this document defines [$insert text node validation steps$] for {{HTMLScriptElement}} to ascertain that text nodes inserted under {{HTMLScriptElement}} elements will satisfy Trusted Types restrictions:
-
-1.  Let |text| be |node|'s [=child text content=].
-1.  Let |convertedText| be the result of running [$Get Trusted Type compliant string$], passing |parent|'s [=relevant global object=] as |global|, {{TrustedScript}} as |expectedType|, |text| as |input| and `"script.text"` as |sink|.
-    If the algorithm threw an error, rethrow the error.
-
-    Note: If Trusted Types are not enforced, this algorithm will return early and return |input| unmodified.
-1.  Set |node|'s data to |convertedText|.
+On setting, the {{HTMLElement/innerText}}, {{Node/textContent}} and {{HTMLScriptElement/text}} IDL attributes perform the regular steps, and then set [=content object=]'s {{script/[[ScriptText]]}} internal slot value with script's [=child text content=].
```
annevk: I think you want to use the given value here and not the script's child text content. I worry somewhat that a mutation event could have modified the actual value. (Did not try to verify.)

koto: Sounds good.

koto:
Merging - this is for scripts only, and does not resolve other elements or setting content attributes yet.

PR #282, opened 2020-07-03
Fix invalid Web IDL syntax

foolip:
With these changes the Web IDL fragments are possible to parse,
but the use of overloading is still in violation of Web IDL:
https://heycam.github.io/webidl/#idl-overloading

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1,8 +1,3 @@
-<!--
-Before editing this document, please see
-https://github.com/w3c/webappsec-trusted-types/blob/master/README.md#spec-changes
--->
-
 <pre class='metadata'>
 Title: Trusted Types
 Shortname: trusted-types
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1,8 +1,3 @@
-<!--
-Before editing this document, please see
-https://github.com/w3c/webappsec-trusted-types/blob/master/README.md#spec-changes
```
foolip, line 3: This link doesn't work so I removed it.

foolip:
I've filed https://github.com/w3c/webappsec-trusted-types/issues/283 for the overloading problem.

foolip:
@mikewest are you the right reviewer for this?

koto:
Hey, sorry, I missed that in my inbox. 

Thanks for the changes - can you generate the spec file too? `npm install` will generate the correct pre-commit hooks that do that (`npm spec` is the actual build target for the spec).

koto:
Isn't this obsoleted by #283?

saschanaz:
I think #296 shouldn't have removed the `exclude` so that crawlers won't bother with them. The suggestion back then probably was to disable IDL index and keep the exclude classes.

foolip:
This is obsolete, I'll open another PR.

PR #291, opened 2020-08-26
Correct return value in example code

jugglinmike:


koto:
Thanks! Can you also run `npm install` and `npm spec` to generate the html file for the spec?

jugglinmike:
Sure thing. That might be good to document in the contribution guidelines. (For anyone reading along, the second command is `npm run spec`.)

Unfortunately, the project's pre-commit hook launches some processes that are too much for my development machine to handle. They seem to be related to testing and building the polyfill, though. It's under that assumption that I've used `--no-verify` to create this new commit.

PR #298, opened 2020-10-22
spec: typo fix

Malvoz:
s/withour/without


Malvoz:

koto:
Thanks!

Issue #190, opened 2019-07-11
Add a target suitable for nodejs.


koto:
That probably means publishing an npm package that 
  * implements the TT API only (equivalent of `api_only.js` entry point) (`TrustedTypes.createPolicy` etc.).
  * enforces restrictions (i.e. hook to prototype methods) on a given document-like and window-like objects - and a enforcement configuration (currently this is implemented as a CSP string).

These can be separate modules, but should all be packaged in one package, such that one can (I'm making up the API shape completely here):
```
import {TrustedTypes} from 'trusted-types/api.js';
import {enforceOn} from 'trusted-types/enforcement.js'; 
enforceOn(document, window, 'default': TrustedTypes.createPolicy('default', ...), violationCallback);
```

For the time being, this should probably be packaged as [CommonJS](https://flaviocopes.com/commonjs/) modules, as ES modules are still [experimental in Node 12](https://nodejs.org/api/esm.html), but we should explore how to migrate to ES modules eventually. It sounds like we might be able to package both variants together : https://nodejs.org/api/esm.html#esm_package_entry_points.

koto:
@engelsdamien you probably would depend on the polyfill package?

koto:
The current node package ([api_only as commonjs](https://github.com/WICG/trusted-types/blob/master/package.json#L5)) unfortunately breaks:

```
$ node
> require('.')
ReferenceError: HTMLIFrameElement is not defined
    at trustedTypesBuilderTestOnly (/usr/local/google/home/koto/dev/trusted-types/dist/cjs/trustedtypes.api_only.js:357:21)
    at Object.<anonymous> (/usr/local/google/home/koto/dev/trusted-types/dist/cjs/trustedtypes.api_only.js:684:5)
    at Module._compile (internal/modules/cjs/loader.js:701:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:712:10)
    at Module.load (internal/modules/cjs/loader.js:600:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:539:12)
    at Function.Module._load (internal/modules/cjs/loader.js:531:3)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:22:18)
```

[...]

PR #299, opened 2020-10-26
Using TT polyfill in nodejs environment

Siegrift:
This is a preliminary PR adding demo showing support of TT polyfill in `nodejs` targets.

The "API only" works even as of now, however there is no support for enforcing mode when the target framework simulates DOM (e.g. Angular, using [domino](https://github.com/fgnass/domino) adapter for server side). This demo, tries to integrate with domino and support enforcing mode (on server side).

You can run the demo with `yarn run-demo`. Currently, the demo breaks when trying to enforce TT with the following error.
```
/home/siegrift/Documents/google/node-target/trusted-types/build/cjs/trustedtypes.api_only.js:1016
  defineProperty(object, name, {
  ^

TypeError: Cannot redefine property: insertAdjacentHTML
```

The reason is that domino creates the [properties on the DOM](https://github.com/fgnass/domino/blob/12bc33d09b3e82f9c71230d0b0496e9d077fc85a/lib/Element.js#L167) as non-configurable.


Siegrift:
This PR updates the code of the polyfill by making the `api_only.js` side effect free and `full.js` contain all of the side effects (attaching TT to window + enforcing). Also exports the enforcer from `api_only.js` so we have full TT support in nodejs.

The TS type definitions also need to change to:
```
// Type definitions for trusted-types 1.0
// Project: https://github.com/WICG/trusted-types
// Definitions by: Jakub Vrana <https://github.com/vrana>
//                 Damien Engels <https://github.com/engelsdamien>
//                 Emanuel Tesar <https://github.com/siegrift>
//                 Bjarki <https://github.com/bjarkler>
// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
// TypeScript Version: 3.1

type FnNames = keyof TrustedTypePolicyOptions;
type Args<Options extends TrustedTypePolicyOptions, K extends FnNames> =
    Parameters<NonNullable<Options[K]>>;

declare global {
    class TrustedHTML {
        private constructor(); // To prevent instantiting with 'new'.
        private brand: true; // To prevent structural typing.
    }

    class TrustedScript {
        private constructor(); // To prevent instantiting with 'new'.
        private brand: true; // To prevent structural typing.
    }

    class TrustedScriptURL {
        private constructor(); // To prevent instantiting with 'new'.
        private brand: true; // To prevent structural typing.
    }

    interface TrustedTypePolicyFactory {
        createPolicy<Options extends TrustedTypePolicyOptions>(
            policyName: string,
            policyOptions?: Options,
        ): Pick<TrustedTypePolicy<Options>, 'name'|Extract<keyof Options, FnNames>>;
        isHTML(value: unknown): value is TrustedHTML;
        isScript(value: unknown): value is TrustedScript;
        isScriptURL(value: unknown): value is TrustedScriptURL;
        readonly emptyHTML: TrustedHTML;
        readonly emptyScript: TrustedScript;
        getAttributeType(tagName: string, attribute: string, elementNs?: string, attrNs?: string): string | null;
        getPropertyType(tagName: string, property: string, elementNs?: string): string | null;
        readonly defaultPolicy: TrustedTypePolicy | null;
        TrustedHTML;
        TrustedScript;
        TrustedScriptURL;
    }

    interface TrustedTypePolicy<Options extends TrustedTypePolicyOptions = TrustedTypePolicyOptions> {
        readonly name: string;
        createHTML(...args: Args<Options, 'createHTML'>): TrustedHTML;
        createScript(...args: Args<Options, 'createScript'>): TrustedScript;
        createScriptURL(...args: Args<Options, 'createScriptURL'>): TrustedScriptURL;
    }

    interface TrustedTypePolicyOptions {
        createHTML?: (input: string, ...arguments: any[]) => string;
        createScript?: (input: string, ...arguments: any[]) => string;
        createScriptURL?: (input: string, ...arguments: any[]) => string;
    }

    interface Window {
        // `trustedTypes` is left intentionally optional to make sure that
        // people handle the case when their code is running in a browser not
        // supporting trustedTypes.
        trustedTypes?: TrustedTypePolicyFactory;
        TrustedHTML: TrustedHTML;
        TrustedScript: TrustedScript;
        TrustedScriptURL: TrustedScriptURL;
        TrustedTypePolicyFactory: TrustedTypePolicyFactory;
        TrustedTypePolicy: TrustedTypePolicy;
    }
}

declare const trustedTypes: TrustedTypePolicyFactory;

declare class TrustedTypesEnforcer {
    constructor(config: TrustedTypeConfig);
    install: () => void
    uninstall: () => void
}

declare class TrustedTypeConfig {
  constructor(
    isLoggingEnabled: boolean,
    isEnforcementEnabled: boolean,
    allowedPolicyNames: string[],
    allowDuplicates: boolean,
    cspString?: string | null
  )
}

// These are the available exports when using the polyfill as npm package (e.g. in nodejs)
export {trustedTypes, TrustedTypesEnforcer, TrustedTypeConfig};
```

Siegrift:
This PR is intended to resolve #190

demo/nodejs-domino/custom.ts
```diff
@@ -0,0 +1,53 @@
...
+html.appendChild(doc.createElement('div'))
+
+console.log(html.innerHTML)
+
+// no fail
+html.children[0].innerHTML = 'string'
+console.log(html.innerHTML)
+
+const config = new TrustedTypeConfig(false, true, ['foo'], false)
+const enforcer = new TrustedTypesEnforcer(config, win)
```
koto, line 33: You could also demonstrate the default policy:

```javascript
const config = new TrustedTypeConfig(false, true, ['foo', 'default'], false)
const enforcer = new TrustedTypesEnforcer(config)
enforcer.install()

const fooPolicy = trustedTypes.createPolicy('foo', { createHTML: (s) => s })

let caught = false;
try {
  html.children[0].innerHTML = 'string' // should fail
} catch (err) {
  caught = true;
}
if (caught) {
  console.log('Caught unsafe write to innerHTML property.')
} else {
  throw new Error('Didn\'t catch unsafe write to innerHTML property!');
}

// Default policy
trustedTypes.createPolicy('default', { createHTML: (s) => s + '-default'});

html.children[0].innerHTML = 'string' // should be rewritten to 'string-default'
console.log(html.children[0].innerHTML);

```

Note how I changed the exception handler too (see other comment).

demo/nodejs-domino/custom.ts
```diff
@@ -0,0 +1,53 @@
...
+enforcer.install()
+const fooPolicy = trustedTypes.createPolicy('foo', { createHTML: (s) => s })
+try {
+  html.children[0].innerHTML = 'string' // should fail
+  throw new Error('Unexpected code path')
```
koto, line 38: This throw makes it confusing, as you don't know in the exception handler below if the `innerHTML` actually succeeded (this is what was breaking the default policy code we talked about).

demo/nodejs-domino/custom.ts
Siegrift: Yep, make sense.

src/enforcer.js
```diff
@@ -158,6 +124,73 @@ export class TrustedTypesEnforcer {
...
+ * @private
+ */
+  getConstructorName_(fn) {
+    if (!this.windowObject.document) return null;
+    return this.windowObject.document.createElement('div').constructor.name ?
+      fn.name :
+      ('' + fn).match(/^\[object (\S+)\]$/)[1];
   }
 
   /**
```
koto, line 195: Can you mention in install() description that now the unknown properties are not hooked, instead of throwing, and why this was needed?

src/enforcer.js
```diff
@@ -258,22 +322,24 @@ export class TrustedTypesEnforcer {
...
+            function(originalFn, ...args) {
+              return that.enforceTypeInScriptNodes_
+                  .bind(that, this, /* checkParent */ false, originalFn)
+                  .apply(that, args);
+            });
+      }
     });
     this.wrapFunction_(
-        insertAdjacentObject,
+        this.insertAdjacentObjectPrototype,
```
koto, line 341: This might be null, should we not skip wrapFunction_ in this case, like above? Also in uninstall().

src/enforcer.js
```diff
@@ -158,6 +124,73 @@ export class TrustedTypesEnforcer {
...
+  }
+
+  /**
+     * Checks whether the value is instanceOf the specific window object.
+     * @param {*} value
+     * @param {!string} winProp
+     * @return {!boolean}
+     * @private
+     */
+  instanceOfDomProperty(value, winProp) {
```
koto, line 158: The convention for private fields and methods is to suffix them with _, still (unlike in TS). Please change here and in e.g. windowObject.

src/enforcer.js
```diff
@@ -463,19 +548,25 @@ export class TrustedTypesEnforcer {
    * @return {*}
    */
   enforceTypeInScriptNodes_(context, checkParent, originalFn, ...args) {
+    if (!this.windowObject.Node || !this.windowObject.document) return;
```
koto, line 550: You already check for Node in one place, when installing this function. In the other you check for Element only. Add a check for Node there, and remove from here. When a hooking function is already installed, I would prefer if it threw when any strange environment issue is detected.

src/enforcer.js
```diff
@@ -541,13 +637,14 @@ export class TrustedTypesEnforcer {
 
   /**
    * Wraps a setter with the enforcement wrapper.
-   * @param {!Object} object The object of the to-be-wrapped property.
+   * @param {Object} object The object of the to-be-wrapped property.
```
koto, line 639: Keep it as non-nullable (and skip calling with null as an object).

src/enforcer.js
```diff
@@ -572,13 +669,18 @@ export class TrustedTypesEnforcer {
    * @param {function(!Function, ...*)} functionBody The wrapper function.
    */
   wrapFunction_(object, name, functionBody) {
+    if (!object) return;
```
koto, line 671: The type system should be covering this case.

src/enforcer.js
```diff
@@ -572,13 +669,18 @@ export class TrustedTypesEnforcer {
...
     const descriptor = getOwnPropertyDescriptor(object, name);
     const originalFn = /** @type function(*):* */ (
         descriptor ? descriptor.value : null);
 
     if (!(originalFn instanceof Function)) {
-      throw new TypeError(
```
koto, line 580: Huh? Does it throw on node? When? isLoggingEnabled is about the TT errors in your application, not about errors in your environment.

src/enforcer.js
```diff
@@ -610,16 +712,16 @@ export class TrustedTypesEnforcer {
    * @private
    */
   wrapSetter_(object, name, type, descriptorObject = undefined) {
-    if (descriptorObject && !isPrototypeOf.call(descriptorObject, object)) {
+    if (!this.windowObject.Node) return;
```
koto, line 714: I'd prefer if we didn't call the function when Node is not available instead.

src/enforcer.js
```diff
@@ -834,34 +987,42 @@ export class TrustedTypesEnforcer {
       console.warn(message, propertyName, context, typeToEnforce, value);
     }
 
+    // This is not available in release Firefox :(
```
koto, line 989: Can't reproduce. Firefox has them enabled since ~2 years ago.

src/polyfill/api_only.js
```diff
@@ -11,60 +11,6 @@
  * @fileoverview Entry point for a polyfill that only defines the types
  * (i.e. no enforcement logic).
  */
-import {trustedTypes, TrustedTypePolicy, TrustedTypePolicyFactory} from
```
koto, line 14: As discussed offline, please create a separate entry point for node, and keep api_only without the enforcer object.

We could also split this in two PRs if you want, one for making enforcer support installation on a passed object, and then the polyfill entry points + demo separately.

koto (changes requested):

src/enforcer.js
```diff
@@ -146,9 +112,9 @@ const TrustedTypePolicy = {};
 export class TrustedTypesEnforcer {
   /**
    * @param {!TrustedTypeConfig} config The configuration for
-   * trusted type enforcement.
+   * @param {!Window=} windowObject object to monkey patch
    */
-  constructor(config) {
+  constructor(config, windowObject) {
```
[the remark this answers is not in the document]
Siegrift: Sure.

src/enforcer.js
```diff
@@ -158,6 +124,73 @@ export class TrustedTypesEnforcer {
...
+          ['insertAdjacentHTML']) ?
+          w.Element.prototype :
+            w.HTMLElement.prototype;
+    })(this.windowObject);
+  }
+
+  /**
+     * Checks whether the value is instanceOf the specific window object.
+     * @param {*} value
+     * @param {!string} winProp
```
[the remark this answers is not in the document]
Siegrift: Done.

src/enforcer.js
```diff
@@ -158,6 +124,73 @@ export class TrustedTypesEnforcer {
...
+    const obj = this.windowObject[winProp];
+
+    return value instanceof obj;
+  }
+
+  /**
+ * Converts an uppercase tag name to an element constructor function name.
+ * Used for property setter hijacking only.
+ * @param {string} tagName
+ * @return {string|null}
```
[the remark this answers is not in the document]
Siegrift: Done.

src/enforcer.js
```diff
@@ -158,6 +124,73 @@ export class TrustedTypesEnforcer {
...
+          this.windowObject.document.createElement(tagName).constructor
+      );
+    }
+  }
+
+  /**
+ * Return object constructor name
+ * (their function.name is not available in IE 11).
+ * @param {Function} fn
+ * @return {string|null}
```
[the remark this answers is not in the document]
Siegrift: Done.

src/enforcer.js
```diff
@@ -158,6 +124,73 @@ export class TrustedTypesEnforcer {
...
+    } else if (!this.windowObject.document) {
+      return null;
+    } else {
+      return this.getConstructorName_(
+          this.windowObject.document.createElement(tagName).constructor
+      );
+    }
+  }
+
+  /**
```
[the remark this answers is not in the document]
Siegrift: Done.

src/enforcer.js
```diff
@@ -158,6 +124,73 @@ export class TrustedTypesEnforcer {
...
+ * Return object constructor name
+ * (their function.name is not available in IE 11).
+ * @param {Function} fn
+ * @return {string|null}
+ * @private
+ */
+  getConstructorName_(fn) {
```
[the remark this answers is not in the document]
Siegrift: Done.

src/enforcer.js
Siegrift: Done.

src/enforcer.js
Siegrift: I verify truthiness in the wrap/restore functions.

src/enforcer.js
Siegrift: Done. (That is a wierd convention though...)

src/enforcer.js
Siegrift: I don't get this part.

>  In the other you check for Element only. Add a check for Node there, and remove from here.

Elaborate pls.

src/enforcer.js
Siegrift: Mhh, I've changed this to `nullable` as it's a bit easier to implement. Do you want me to reveert that and make this non-nullable?

src/enforcer.js
Siegrift: True. I don't want to throw here because it will likely break in node (I think even domino breaks here). 

Not sure what to do though? Silent return?

src/enforcer.js
Siegrift: So we get rid of this feature test and just `this.windowObject.SecurityPolicyViolationEvent`?

koto: We still need it for, e.g. IE.

src/polyfill/api_only.js
Siegrift: Done.

src/enforcer.js
```diff
@@ -158,6 +124,73 @@ export class TrustedTypesEnforcer {
...
+    if (!this.windowObject || typeof this.windowObject !== 'object') {
+      throw new Error(
+          // eslint-disable-next-line
+          'The polyfill expects a global `window` object or emulated `window-like` object passed to the enforcer as second argument'
+      );
+    }
+
+    // In IE 11, insertAdjacent(HTML|Text) is on HTMLElement prototype
+    this.insertAdjacentObjectPrototype = ((w) => {
+      if (!w.Element || !w.HTMLElement) return null;
```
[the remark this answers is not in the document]
Siegrift: Done.

Siegrift:
Getting back to our offline discussion about requiring some common DOM symbols (Node, Element...) in the window like object:

I propose that we require these objects: `Element`, `HTMLElement`, `Document`, `HTMLDocument`, `Node` and `document`.
- Each of these has it's specific sinks that we guard on it's prototype. One exception is `document` which we need to to create DOM node on a few places to emulate TT behaviour.
- We would still check if the sink exists on the prototype, but would expect the prototype itself to exist. E.g. we would check `'after' in this.windowObject_.Element.prototype`, but we assume `Element.prototype` exists.
- Examples of non-required DOM properties include `DOMParser`, `ShadowRoot`, `Range` (and every other except from the whitelist above)

Do you agree with this suggestion?

koto:
That sounds good. Assume that the root object has `Element`, `HTMLElement`,`Document`, `HTMLDocument`, `Node` and `document`, and test for existence of anything else you need to hook on.

src/enforcer.js
Siegrift: Done.

src/enforcer.js
Siegrift: Done.

src/enforcer.js
Siegrift: Done.

src/enforcer.js
Siegrift: Done.

src/enforcer.js
Siegrift: Done.

src/enforcer.js
Siegrift: Done.

Siegrift:
Slight change of plans:
- Let's make `HTMLElement` optional, because it's not in domino and we don't really need it that much.

Latest code should be aligned with that we talked about back in the old days :)))

src/enforcer.js
Siegrift: Reverted back to throwing an error.

koto:

koto:
The tests fail on Firefox (see https://travis-ci.com/github/w3c/webappsec-trusted-types/builds/201779040)

Siegrift:
Mhh, the tests pass for me locally when run on headless firefox. What command you execute on CI exactly?

koto:
All is configured in `.travis.yml`. Travis images currently have Firefox 63. I reproduced the error on Browserstack with Firefox 63.0.0 as well. 

You can reproduce the failures locally by downloading a Firefox 63 binary and setting a [FIREFOX_BIN]( https://github.com/karma-runner/karma-firefox-launcher#environment-variables) variable for `npm test` command.

I've partially debugged this - it seems like sometime between 63 and current stable Firefox moved the `write` function from `HTMLDocument` to `Document` (like other browsers). The feature tests are incomplete for the uninstallation `Document.write` branch.

```

-    if (getOwnPropertyDescriptor(
+    if (this.windowObject_.Document && getOwnPropertyDescriptor(
         this.windowObject_.Document.prototype, 'write')
     ) {
       this.restoreFunction_(this.windowObject_.Document.prototype, 'write');
     } else if (this.windowObject_.HTMLDocument && getOwnPropertyDescriptor(
         this.windowObject_.HTMLDocument, 'write')
     ) {
-      this.restoreFunction_(HTMLDocument.prototype, 'write');
+      this.restoreFunction_(this.windowObject_.HTMLDocument.prototype, 'write');
     }
```

This fixes some of the issues, but there's still something about the `setAttribute` function.

src/polyfill/api_only.js
```diff
@@ -40,31 +40,31 @@ function setupPolyfill() {
...
-    'isScriptURL': tt.isScriptURL,
-    'isScript': tt.isScript,
-    'createPolicy': tt.createPolicy,
-    'getAttributeType': tt.getAttributeType,
-    'getPropertyType': tt.getPropertyType,
-    'getTypeMapping': tt.getTypeMapping,
-    'emptyHTML': tt.emptyHTML,
-    'emptyScript': tt.emptyScript,
-    '_isPolyfill_': true,
+    isHTML: trustedTypes.isHTML,
```
koto, line 43: These properties have to be quoted, otherwise the compiler will rename them, and e.g. `window.trustedTypes.isHTML` will not exist.

Siegrift: Oh, thanks for noticing. It was caused by autoformat :/ and I didn't notice it.

dist/spec/index.html
```diff
@@ -26,6 +26,7 @@
  *   - .figure and .caption (HTML4) and figure and figcaption (HTML5)
```
koto, line 26: Can you revert the changes to this file?

package.json
```diff
@@ -2,8 +2,8 @@
   "name": "trusted-types",
   "version": "1.1.1",
   "description": "Polyfill for the Trusted Types",
-  "main": "build/cjs/trustedtypes.api_only.js",
-  "module": "build/esm/trustedtypes.api_only.js",
+  "main": "build/cjs/trustedtypes.full_api_only.js",
```
koto, line 5: Nit: perhaps `trustedtypes.module.js`, since this is to be used like a ES module?

src/enforcer.js
```diff
@@ -173,43 +222,63 @@ export class TrustedTypesEnforcer {
...
+    if (getOwnPropertyDescriptor(
+        this.windowObject_.Document.prototype, 'write'
+    )) {
       // Chrome
-      this.wrapWithEnforceFunction_(Document.prototype, 'write',
+      this.wrapWithEnforceFunction_(this.windowObject_.Document.prototype,
+          'write',
           TrustedTypes.TrustedHTML, 0);
-    } else {
+    } else if (this.windowObject_.HTMLDocument &&
```
koto, line 258: You can drop this check, since we assume that `HTMLDocument` is always available.

src/enforcer.js
```diff
@@ -225,25 +294,44 @@ export class TrustedTypesEnforcer {
...
+    this.restoreFunction_(
+        this.windowObject_.Element.prototype, 'setAttribute');
+    this.restoreFunction_(
+        this.windowObject_.Element.prototype, 'setAttributeNS');
+
+    if (getOwnPropertyDescriptor(
+        this.windowObject_.Document.prototype, 'write')
+    ) {
+      this.restoreFunction_(this.windowObject_.Document.prototype, 'write');
+    } else if (this.windowObject_.HTMLDocument && getOwnPropertyDescriptor(
```
koto, line 320: Same, you can drop this check.

src/enforcer.js
```diff
@@ -225,25 +294,44 @@ export class TrustedTypesEnforcer {
...
+        this.windowObject_.HTMLDocument, 'write')
```
koto, line 321: `this.windowObject_.HTMLDocument.prototype` (this is what broke the tests)

Siegrift: Yep, verified with Firefox binary 63 locally.

src/enforcer.js
Siegrift: HTMLDocument doesn't exist in domino.

package.json
Siegrift: Sure.

dist/spec/index.html
Siegrift: I think I have old version of bikeshed (or whatever it's called).

koto:
Merged, thanks a lot!

PR #308, opened 2021-02-17
Fix remaining Web IDL issues

foolip:
The default value {} is required by Web IDL.

The exclude attributes will cause the Reffy to not extract the
overloaded IDL into webref. These attributes were previously removed
in https://github.com/w3c/webappsec-trusted-types/pull/297.


sideshowbarker:
@koto This one seems to just be waiting on review

PR #346, opened 2021-07-19
fix typos

0xedward:
fixing some typos I found while reading the spec


koto:
That's a lot of typos. Sorry for that and thanks for fixing!

PR #361, opened 2022-03-10
Fix minor inconsistencies

Siegrift:


spec/index.bs
```diff
@@ -629,7 +629,7 @@ initially empty.
 
     <div class="example" id="get-attribute-type-example">
     <xmp highlight="js">
-    trustedTypes.getAttributeType('script', 'SRC'); // "TrustedScriptURL"
+    trustedTypes.getAttributeType('script', 'src'); // "TrustedScriptURL"
```
Siegrift, line 632: The spec says it should be lowercased.

spec/index.bs
```diff
@@ -807,9 +807,9 @@ the default policy is applied.
 // Content-Security-Policy: trusted-types default; require-trusted-types-for 'script'
 
 trustedTypes.createPolicy('default', {
-  createScriptURL: (s, type, sink) => {
+  createScriptURL: (value, type, sink) => {
```
Siegrift, line 810: I've renamed this to `value` based on the edits of the text above.

PR #404, opened 2024-01-15
Collect HTML injection sinks and DOM XSS injection sinks under XSS injection sinks

### LINKED ISSUE #383: Should `require-trusted-types-for` support trusted scripts and trusted script URLs?
https://w3c.github.io/trusted-types/dist/spec/#require-trusted-types-for-csp-directive contains a note that currently only the enforcement of trusted types for DOM XSS injection sinks is supported. The corresponding ABNF form reflects that.

However trusted scripts (https://w3c.github.io/trusted-types/dist/spec/#trusted-script) and trusted script urls (https://w3c.github.io/trusted-types/dist/spec/#trused-script-url) are specified.

mbrodesser-Igalia:
Allowed clarifying that all XSS injection sinks are covered by the
"trusted-types-sink-group" named 'script'.

Closes #383

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -67,10 +67,10 @@ if `aString` contains untrusted data, `foo[bar] = aString` is a statement
 that potentially can trigger a vulnerability, depending on a value
 of `foo` and `bar`.
 
-This document focuses on preventing DOM-Based Cross-Site Scripting
+This document focuses on preventing Cross-Site Scripting
 that occurs when attacker-controlled data reaches
-[[#dom-xss-injection-sinks]], as that eventually causes execution of the
-script payload controlled by the attacker. DOM XSS is prevalent in the
+[[#xss-injection-sinks]], as that eventually causes execution of the
+script payload controlled by the attacker. XSS is prevalent in the
 web applications as there are over 60 different
 injection sinks (e.g. `Element.innerHTML`, or `Location.href` setters).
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -67,10 +67,10 @@ if `aString` contains untrusted data, `foo[bar] = aString` is a statement
 that potentially can trigger a vulnerability, depending on a value
 of `foo` and `bar`.
 
-This document focuses on preventing DOM-Based Cross-Site Scripting
```
koto, line 70: Please keep the DOM-Based XSS or DOM XSS in the spec text (and the link fragments). Trusted Types only covers DOM XSS (Type 0 XSS). The name is a bid misleading, but DOM XSS covers all sinks available to client side code (i.e. DOM API sinks, but also eval, Location etc). This is as opposed to server-side XSSes (Type 1 and 2).

This applies to all mentions below.

spec/index.bs
```diff
@@ -205,14 +205,17 @@ it's not easy to distinguish one from the other.
 This document organizes the injection sinks into groups, based on the
 capabilities that sinks in a given group have. [=Enforcement=] for groups is controlled via <a>trusted-types-sink-group</a> values.
 
-### HTML injection sinks ### {#html-injection-sinks}
```
koto, line 208: Perhaps it might be easier to just merge those two existing sections, and separate them once there is another use case specified?

https://github.com/w3c/trusted-types/pull/407

koto (changes requested):

spec/index.bs
mbrodesser-Igalia: Agree. Thanks for the explanation.

spec/index.bs
mbrodesser-Igalia: Agree.

Issue #398, opened 2024-01-15
Defer `fromLiteral`?


lukewarlow:
While the hypothetical use case for `fromLiteral` is explained it's unclear if that actually translates into real world usage of this API.   

It also suffers from implementability issues, the underlying TemplateMap functionality isn't implemented inside of JavaScriptCore (afaict), and was only implemented in V8 when this function was implemented inside of Chromium. It never shipped inside of Chromium despite having an intent to ship and this seems to be as a result of technical difficulties in the implementation. I imagine similar issues may exist in Gecko.

There's also no way to properly polyfill this without exposing a potential bypass (no userland way to detect a true template literal).

With all this in mind it would make sense to defer this to v2 of the spec, that way there's more time for something like https://tc39.es/proposal-array-is-template-object/ to progress to make both polyfilling and implementation easier.

cc @mbrodesser-Igalia @bkardell

lukewarlow:
There's also currently spec issues with this functionality see https://github.com/w3c/trusted-types/issues/374

caridy:
At salesforce, we are ok deferring `fromLiteral`.

mbrodesser-Igalia:
Wondering if deferring `fromLiteral` would prevent web-developers from enforcing to forbid trusted type policies.

That's because `Content-Security-Policy: trusted-types 'none'; require-trusted-types-for 'script'`allows the usage of `fromLiteral`. Hence simplifying to transition away from trusted types policies to non-XSS APIs instead. The latter is mentioned as a goal in https://github.com/mozilla/standards-positions/issues/20#issuecomment-1783279722.

Here, experience from people having actually applied TT could be beneficial.

lukewarlow:
I think it's definitely a valuable addition to the API  in terms of simplifying certain code paths. But I think given Google and other's success without it (it's not shipped in Chrome) + the implementation difficulties it's probably worth defering.

I agree the end goal is to just do `trusted-types;` (or `trusted-types 'none'`) and just not using XSS APIs. I think the reality of that will depend on the santizer API and whether it effectively counts as a trusted sink. I'm unaware how else you'd use web components for example without having at least *some* XSS sink usage.

mbrodesser-Igalia:
> I think it's definitely a valuable addition to the API in terms of simplifying certain code paths. But I think given Google and other's success without it (it's not shipped in Chrome) + the implementation difficulties + lack of polyfillability (which I think is quite important) it's probably worth defering.
> 

Makes sense.

[...]

Issue #374, opened 2022-09-16
TrustedHTML.fromLiteral is exposed in workers but assumes the current global is a Window


Ms2ger:
In https://w3c.github.io/trusted-types/dist/spec/#abstract-opdef-create-a-trusted-type-from-literal , TrustedHTML is special-cased to do

> Let templateNode be the results of [creating an element](https://dom.spec.whatwg.org/#concept-create-element) given "template", the [HTML namespace](https://infra.spec.whatwg.org/#html-namespace) and [current global object](https://html.spec.whatwg.org/multipage/webappapis.html#current-global-object)'s [associated Document](https://html.spec.whatwg.org/multipage/window-object.html#concept-document-window).

but the current global could be a worker global.

PR #405, opened 2024-01-15
Remove `fromLiteral` from spec.

lukewarlow:
Addresses #398 

A follow up draft PR to add it back in can be made afterwards, and merged once Level 1 is versioned.


lukewarlow:
Draft pending any discussion in the linked issue

lukewarlow:
The follow up should address https://github.com/w3c/trusted-types/issues/374 - either Matching Chrome's flagged behaviour or simply not exposing in Workers at all (I have a slight preference for the latter)

lukewarlow:
cc @koto @mbrodesser-Igalia @bkardell 

I think based on our discussion the exact how for implementing fromLiteral is still up in the air enough to warrant deffering from v1, with the understanding that it does have value and the issues should be worked out for it to be added back in the future. On that basis this PR removes it from the current draft spec.

lukewarlow:
https://github.com/web-platform-tests/wpt/pull/44350 - Test PR to mark as tentative

PR #408, opened 2024-01-17
Add use case for safely transitioning websites to not use any DOM XSS injection sinks

mbrodesser-Igalia:
As mentioned at
<https://github.com/mozilla/standards-positions/issues/20#issuecomment-1783279722>.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -95,6 +95,11 @@ monitor and review.
 *   Encourage a design in which security decisions are
     encapsulated within a small part of the application.
 
+*   Enable transitioning existing websites to call alternative save methods to
+    [[#dom-xss-injection-sinks]]. Potentially with support of the <a
+    http-header>Content-Security-Policy-Report-Only</a> HTTP response header
+    field.
+
 *   Reduce security review surface for complex web application
     codebases.
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -95,15 +95,10 @@ monitor and review.
 *   Encourage a design in which security decisions are
     encapsulated within a small part of the application.
 
-*   Enable transitioning existing websites to call alternative save methods to
-    [[#dom-xss-injection-sinks]]. Potentially with support of the <a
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -95,6 +95,11 @@ monitor and review.
 *   Encourage a design in which security decisions are
     encapsulated within a small part of the application.
 
+*   Enable transitioning existing websites to call alternative save methods to
```
koto, line 98: That seems already captured by the existing goals, no? What is mentioned here is just one of technical means to "minimize the likelihood of client-side vulnerabilities", and "reduce security review surface".

mbrodesser-Igalia: Yes. The technical means and its more explicit, since it mentions transitioning existing websites.
TBH this might be obvious to someone very familiar with the API and the topic, for others not.

Feel free to disagree though and reject this change.

mbrodesser-Igalia: "Enable iteratively transitioning" would be even clearer from my point of view.

The only other sections where that information seems correctly placed is the introduction or the use cases section.

Identifying that real-world use case took, at least me, some time.

mbrodesser-Igalia: WDYT about changing: "Allow the detection of vulnerabilities" to "Allow the usability-preserving detection of". And adding a use case like

'A website uses DOM XSS injection sinks. The website-developer adds trusted types to it and monitors violations by using the Content-Security-Policy-Report-Only header field. Violations are iteratively fixed by refactoring the code to use only save methods. Afterwards, no DOM XSS injection sinks are called anymore. Hence, no trusted types are required anymore. The developer switches the report-only mode off and disables trusted type policies with the `require-trusted-types-for directive`. The website's functionality was never impaired during the refactorings.'

koto: I like that. The only change I'd suggest would be "save" -> "safe".

mbrodesser-Igalia: Added a patch for it with that fixed and disabling trusted type policies slightly more explained.

koto (approved):

PR #416, opened 2024-01-23
Add WPT block to spec to display test link

lukewarlow:
This includes a link to the tests and a block showing current pass rate.

This will need to be maintained as new tests are added (or removed) but it feels a useful addition.

Screenshot 2024-01-23 at 12 24 42


lukewarlow:
This can be changed to split the wpt blocks up into the spec itself so you can get links to the tests covering a specific algorithm or section. I think this could be valuable to evaluate test coverage more easily.

PR #417, opened 2024-01-23
Stringify input when trusted type

### LINKED ISSUE #413: Process value with default policy missing step?
https://w3c.github.io/trusted-types/dist/spec/#process-value-with-a-default-policy-algorithm - takes a TrustedType or string but the algorithm it calls only takes a string. Shouldn't trusted type be stringified before passing?

Apologies if this is some sort of implied step in algorithms I'm not aware of.

lukewarlow:
Fixes #413

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -990,7 +990,7 @@ Given a {{TrustedType}} type (|expectedType|), a [=realm/global object=] (|globa
 1.  Let |convertedInput| be the result of executing [$Create a
     Trusted Type$] algorithm, with the following arguments:
     *  |defaultPolicy| as |policy|
-    *  |input| as |value|
+    *  stringified |input| as |value|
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -990,7 +990,7 @@ Given a {{TrustedType}} type (|expectedType|), a [=realm/global object=] (|globa
 1.  Let |convertedInput| be the result of executing [$Create a
     Trusted Type$] algorithm, with the following arguments:
     *  |defaultPolicy| as |policy|
-    *  |input| as |value|
+    *  |input| as |value|, stringifying if |input| is a {{TrustedType}}
```
koto, line 993: You could also just say `stringified |input| as |value|`, stringifying a string is a noop.

koto (approved):

Issue #359, opened 2022-02-15
Maybe enforce Trusted Types in XSL's xsl:text


shhnjk:
Currently, there is no Trusted Types enforcement on `<xsl:text>` in XSL document.
Found by [Alex](https://twitter.com/insertScript)
https://twitter.com/kinugawamasato/status/1493641462776360961

```
<script>
  let attackerControlledString = "<img src=x onerror=alert(origin)>";
  const doc = document.implementation.createHTMLDocument();
  const xslt = document.createElementNS("http://www.w3.org/1999/XSL/Transform","xsl:stylesheet");
  xslt.setAttribute("xmlns:xsl","http://www.w3.org/1999/XSL/Transform");
  const template = document.createElementNS("http://www.w3.org/1999/XSL/Transform","xsl:template");
  template.setAttribute("match","/");
  const output = document.createElementNS("http://www.w3.org/1999/XSL/Transform","xsl:output");
  output.setAttribute("method","html");
  xslt.appendChild(output);
  const text = document.createElementNS("http://www.w3.org/1999/XSL/Transform","xsl:text");
  text.textContent = attackerControlledString;
  text.setAttribute("disable-output-escaping","yes");
  template.appendChild(text);
  xslt.appendChild(template);
  const processor = new XSLTProcessor();
  processor.importStylesheet(xslt);
  const fragment = processor.transformToFragment(doc,document);
  document.body.appendChild(fragment);
</script>
```

shhnjk:
The actual root cause of XSS is due to `disable-output-escaping` attribute set to `yes`, which will disable HTML escaping. The only easy way to mitigate this bug seems to be disabling `disable-output-escaping` attribute under Trusted Types enforcement. Per [this document](https://developer.mozilla.org/en-US/docs/Web/XSLT/Common_errors#missing_features), Firefox doesn't even support `disable-output-escaping`.

koto:
`XSLTProcessor.importStylesheet` could also be a sink in this case, if the stylesheet nodes have a capability of causing  script execution.

annevk:
Yeah, XSLT entry points should be blocked. We shouldn't support XSLT in detail.

mbrodesser-Igalia:
How would the entry points, e.g. `XSLTProcessor.importStylesheet` (https://developer.mozilla.org/en-US/docs/Web/API/XSLTProcessor/importStylesheet), be blocked?

[...]

Issue #260, opened 2020-02-19
Document.write and such as names are misleading


annevk:
The `Document` class does not have a static `write` property. I think we should make this `Document.prototype.write` or at least something that does not suggest a misleading thing (and which well might conflict in the future).

Via https://github.com/heycam/webidl/pull/841#issuecomment-586453624.

koto:
Thanks, that's a good point. We're taking the data from the context object in steps 1.1 and 1.2 [here](https://w3c.github.io/webappsec-trusted-types/dist/spec/#!trustedtypes-extended-attribute), so it should be easy. There is still a potential collision as we're using the DOM local names for Elements for simplification.

annevk:
I guess elements will always be lowercase? However, you probably want to distinguish SVG script from HTML script?

koto:
There's probably no conflict possibility for static vs regular operations. In [WebIDL 2.5.3](https://heycam.github.io/webidl/#idl-operations):

> The identifier of a static operation also must not be the same as the identifier of a regular operation defined on the same interface

There's a similar restriction for attributes:

> The identifier of an attribute must not be the same as the identifier of another interface member defined on the same interface. The identifier of a static attribute must not be "prototype".

I looked into [our implementation](https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/trustedtypes/trusted_types_util.cc?l=93&rcl=8cc471ecb167c819f29b10a5a3e9962ef5054d64), and it doesn't seem like we have the information to distinguish between prototype / static readily avaliable in the exception state. We actually (contrary to the spec) always use the constructor name, so:

```
data:text/html,<meta http-equiv="content-security-policy" content="require-trusted-types-for 'script'"><script>addEventListener('securitypolicyviolation', (e) => alert(e.sample));s=document.createElement('iframe');s.srcdoc = 'aaa'</script>
```

alerts `HTMLIframeElement.srcdoc`

It sounds like this might actually be a good outcome - if we always use the constructor name, that suggests to the authors that the sample contains the information about the sink in IDL (which is what the checks will hook on to), regardless on whether it's static or on the prototype.

annevk:
I didn't realize IDL forbids conflicts, but I don't think it's a good idea to standardize on this rather misleading naming either.

cc @domenic

domenic:
+1, as a web developer I find it supremely confusing when browser developers start talking about static methods and properties (like Document.write or HTMLIFrameElement.srcdoc) which do not exist.

[...]

PR #418, opened 2024-01-23
Rewrote DOM integration

### LINKED ISSUE #401: Is the `[[ScriptURL]]` slot needed?
https://w3c.github.io/trusted-types/dist/spec/#slots-with-trusted-values - say
> Figure out if we can drop [[[ScriptURL]]](https://w3c.github.io/trusted-types/dist/spec/#dom-htmlscriptelement-scripturl-slot) slot after IDL + DOM changes.

So we need to figure out if it can indeed be dropped as a result of any other changes that have been made in this spec (or if there's work needed to achieve it?).

koto:
This adds an entry point algorithm to call from DOM's append, change, and replace algorithms for Attr nodes.

https://github.com/whatwg/dom/issues/789
Closes #401.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1101,8 +1101,8 @@ Given an {{HTMLScriptElement}} (|script|), this algorithm performs the following
 
     If the algorithm threw an error, rethrow the error.
 
-## <dfn-abstract-op export>Validate attribute mutation</dfn> ## {#validate-attribute-mutation}
-Given {{Attr}} |attribute|, {{Element}} |element| and {{TrustedType}} or a string |newValue|, this algorithm performs the following steps:
+## Get Trusted Types-compliant attribute value ## {#validate-attribute-mutation}
+To <dfn abstract-op export>get Trusted Types-compliant attribute value</dfn> on {{Attr}} |attribute| with {{Element}} |element| and {{TrustedType}} or a string |newValue|, perform the following steps:
 
 1. If |element| has an <a spec="webidl" lt="attribute">IDL attribute</a> that [=reflects=] a content attribute with |attribute|'s [=Attr/local name=], let |idlAttribute| be that IDL attribute. Otherwise, return stringified |newValue|.
 1. If {{StringContext}} IDL extended attribute does not appears on |idlAttribute|, return stringified |newValue|.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1110,7 +1110,7 @@ To <dfn abstract-op export>get Trusted Types-compliant attribute value</dfn> on 
      * {{StringContext}} identifier as |expectedType|
      * |newValue| as |input|
      * |element|'s <a>relevant global object</a> as |global|
-     * the result of [=concatenating=] « |element|'s [=Element/local name=], |attribute|'s |localName| » with `"."` as a |separator| as |sink|
+     * the result of [=concatenating=] « |element|'s [=Element/local name=], |attribute|'s [=Attr/local name=] » with `"."` as a |separator| as |sink|
      * 'script' as |sinkGroup|
 
    If the algorithm threw an error, rethrow the error.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1091,18 +1091,6 @@ Given a {{TrustedType}} type (|expectedType|), a [=realm/global object=] (|globa
 
 Given an {{HTMLScriptElement}} (|script|), this algorithm performs the following steps:
 
-1.  If |script| does not have a <{script/src}> content attribute, set its {{HTMLScriptElement/[[ScriptURL]]}} internal slot value to `null`.
-
-1.  Otherwise, if |script|'s {{HTMLScriptElement/[[ScriptURL]]}} internal slot value is not equal to its <{script/src}> attribute value,
-    set |script|'s {{HTMLScriptElement/[[ScriptURL]]}} to the result of executing [$Get Trusted Type compliant string$], with the following arguments:
-      * {{TrustedScriptURL}} as |expectedType|,
-      * |script|'s {{Document}}'s [=relevant global object=] as |global|,
-      * |script|'s <{script/src}> attribute value as |input|,
-      * `HTMLScriptElement src` as |sink|,
-      * `'script'` as |sinkGroup|.
-
-    If the algorithm threw an error, rethrow the error and abort further steps.
-
 1.  If |script|'s {{HTMLScriptElement/[[ScriptText]]}} internal slot value is not equal to its [=child text content=],
     set |script|'s {{HTMLScriptElement/[[ScriptText]]}} to the result of executing [$Get Trusted Type compliant string$], with the following arguments:
       * {{TrustedScriptURL}} as |expectedType|,
@@ -1112,6 +1100,20 @@ Given an {{HTMLScriptElement}} (|script|), this algorithm performs the following
       * `'script'` as |sinkGroup|.
 
     If the algorithm threw an error, rethrow the error.
+
+## <dfn-abstract-op export>Validate attribute mutation</dfn> ## {#validate-attribute-mutation}
+Given {{Attr}} |attribute|, {{Element}} |element| and {{TrustedType}} or a string |newValue|, this algorithm performs the following steps:
+
+1. If |element| has an <a spec="webidl" lt="attribute">IDL attribute</a> that [=reflects=] a content attribute with |attribute|'s [=Attr/local name=], let |idlAttribute| be that IDL attribute. Otherwise, return stringified |newValue|.
+1. If {{StringContext}} IDL extended attribute does not appears on |idlAttribute|, return stringified |newValue|.
+1. Set |value| to the result of executing [$Get Trusted Type compliant string$] with the following arguments:
+     * {{StringContext}} identifier as |expectedType|
+     * |newValue| as |input|
+     * |element|'s <a>relevant global object</a> as |global|
+     * the result of [=concatenating=] the list &laquo; |element|'s [=Element/local name=], |attribute|'s |localName| &raquo; with `"."` as a |separator| as |sink|
+     * 'script' as |sinkGroup|
+
+   If the algorithm threw an error, rethrow the error.
 
 # Integrations # {#integrations}
 
@@ -1245,14 +1247,7 @@ partial interface Document {
 
 #### Slots with trusted values #### {#slots-with-trusted-values}
 
-Issue: Figure out if we can drop {{HTMLScriptElement/[[ScriptURL]]}} slot after IDL + DOM changes.
-
 This document modifies {{HTMLScriptElement}}s. Each script has:
-
-: <dfn for="HTMLScriptElement" lt="[[ScriptURL]]" attribute>`[[ScriptURL]]`</dfn> internal slot.
-::  A string, containing the URL to execute the script from
-    that was set through a {{StringContext}} compliant sink. Equivalent to
-    {{HTMLScriptElement/src}} attribute value. Initially null.
 
 : <dfn for="HTMLScriptElement" lt="[[ScriptText]]" attribute>`[[ScriptText]]`</dfn> internal slot.
 ::  A string, containing the body of the script to execute that was set
@@ -1283,12 +1278,6 @@ On setting the {{HTMLElement/innerText}}, {{Node/textContent}} and {{HTMLScriptE
 1. Set {{HTMLScriptElement/[[ScriptText]]}} internal slot value to the stringified attribute value.
 1. Perform the usual attribute setter steps.
 
-
-On setting the {{HTMLScriptElement/src}} IDL attribute, execute the following algorithm:
-
-1. Set {{HTMLScriptElement/[[ScriptURL]]}} internal slot value to the stringified attribute value.
-1. Perform the usual attribute setter steps.
-
 #### Slot value verification #### {#slot-value-verification}
 
 The first few steps of the [=prepare the script element=] algorithm are modified as follows:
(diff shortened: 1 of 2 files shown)


spec/index.bs
```diff
@@ -1113,6 +1101,20 @@ Given an {{HTMLScriptElement}} (|script|), this algorithm performs the following
 
     If the algorithm threw an error, rethrow the error.
 
+## <dfn-abstract-op export>Validate attribute mutation</dfn> ## {#validate-attribute-mutation}
+Given {{Attr}} |attribute|, {{Element}} |element| and {{TrustedType}} or a string |newValue|, this algorithm performs the following steps:
+
+1. If |element| has an <a spec="webidl" lt="attribute">IDL attribute</a> that [=reflects=] a content attribute with |attribute|'s [=Attr/local name=], let |idlAttribute| be that IDL attribute. Otherwise, return stringified |newValue|.
```
annevk, line 1107: This seems super magical. It's also missing a namespace check.

We also can't say "stringified". That needs to be defined.

spec/index.bs
```diff
@@ -1113,6 +1101,20 @@ Given an {{HTMLScriptElement}} (|script|), this algorithm performs the following
...
+1. If {{StringContext}} IDL extended attribute does not appears on |idlAttribute|, return stringified |newValue|.
+1. Set |value| to the result of executing [$Get Trusted Type compliant string$] with the following arguments:
+     * {{StringContext}} identifier as |expectedType|
+     * |newValue| as |input|
+     * |element|'s <a>relevant global object</a> as |global|
```
annevk, line 1112: Do we want this or the element's node document's global? Is there test coverage for this? (Gecko mutates the global when moving elements across documents which is against the specification.) I think in another case we decided to use the element's node document for cases like this, not because of Gecko, but because it would be more likely what is expected.

cc @domenic

spec/index.bs
```diff
@@ -1113,6 +1101,20 @@ Given an {{HTMLScriptElement}} (|script|), this algorithm performs the following
...
+     * the result of [=concatenating=] the list &laquo; |element|'s [=Element/local name=], |attribute|'s |localName| &raquo; with `"."` as a |separator| as |sink|
```
annevk, line 1113: Nit: Use the code points directly rather than entities. Makes it easier to see what is going on. Also no need to prefix with "the list".

annevk, line 1113: Also, why is it |localName| for attribute, that seems wrong?

annevk (commented):
I haven't restudied the wider context, but I have some comments even without context. Hope they are helpful.

spec/index.bs
koto: > This seems super magical

This mirrors https://w3c.github.io/trusted-types/dist/spec/#dom-trustedtypepolicyfactory-getattributetype, in general we decided to encode the controls at IDL layer, so if not reflection, we'd have to use some other mapping mechanism. We could just have a table of attr names + elements ->  type instead, but it feels wrong to duplicate what's already in IDL. An advice would be a great help here.

> Missing a namespace check

The reflection is only [defined for namespace-less attributes](https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#reflected-content-attribute-name), not sure why.

> "stringified" needs to be defined

It is defined? All [TrustedType objects](https://w3c.github.io/trusted-types/dist/spec/#trusted-html) have [stringifiers](https://webidl.spec.whatwg.org/#idl-stringifiers). Do you mean we shoudl avoid the colloquial term "stringified" here and replace it with the explicit algorithm, e.g. branching on a type?

spec/index.bs
koto: In what sense? This is only used in the violation report (i.e. it's not a bypass if we miss a namespace).

koto: nvm, fixed :)

spec/index.bs
annevk: Reflection is not in IDL however. That's only the case in (some?) browsers.

annevk: And yeah, you need to define what stringification means. A naïve implementation could use it as an excuse to execute JavaScript, for instance. (Or maybe that's what you want?)

spec/index.bs
koto: I think it's fine to use the global from the node's document instead, but can you give me an example of when those two globals differ when this algorithm is called? Is it only in Gecko? I don't think we have tests currently for moving nodes across documents and calling setAttribute(), this would help us add a case for this.

annevk: When you move a node from another document its own global will be that other document's global (except in Gecko), but it's node document's global will be that of the new document its been inserted into.

spec/index.bs
koto: Removed IDL reflection. The stringification is now defined, but needs to be applied throughout this spec in other places too. PTAL.

spec/index.bs
```diff
@@ -1113,6 +1101,38 @@ Given an {{HTMLScriptElement}} (|script|), this algorithm performs the following
 
     If the algorithm threw an error, rethrow the error.
 
+## Get Trusted Types-compliant attribute value ## {#validate-attribute-mutation}
+To <dfn abstract-op export>get Trusted Types-compliant attribute value</dfn> on {{Attr}} |attribute| with {{Element}} |element| and {{TrustedType}} or a string |newValue|, perform the following steps:
+
+1. Let |expectedType| be null.
+1. If |attribute|'s <a for="Attr">namespace</a> is null or is a <a>HTML namespace</a>, find the row in the following table, where |element| is in the first column and |attribute|'s <a for="Attr">local name</a> is an <a>ASCII case-insensitive</a> match for any the string in the
```
annevk, line 1108: The check for the HTML namespace here is wrong.

The case-insensitive match seems wrong. Especially if as you say this is IDL-backed in some implementations. Can you point to a test case that exercises that code?

spec/index.bs
```diff
@@ -1113,6 +1101,38 @@ Given an {{HTMLScriptElement}} (|script|), this algorithm performs the following
...
+
+    <table>
+      <thead>
+        <tr><th>Element<th>Attribute local name</th><th>TrustedType<th>
+      <tbody>
+        <tr><td>{{HTMLIFrameElement}}<td>"srcdoc"<td>{{TrustedHTML}}
+        <tr><td>{{HTMLEmbedElement}}<td>"src"<td>{{TrustedScriptURL}}
+        <tr><td>{{HTMLScriptElement}}<td>"src"<td>{{TrustedScriptURL}}
+        <tr><td>{{HTMLObjectElement}}<td>"data"<td>{{TrustedScriptURL}}
+        <tr><td>{{HTMLObjectElement}}<td>"codebase"<td>{{TrustedScriptURL}}
```
annevk, line 1119: I guess this list doesn't include some elements/attributes because of the cross-document thingy. And still includes embed/object due to that being the status quo.

koto: The list mirrors the functionality on the IDL layer that's currently in the spec. It doesn't include the XSL-related things (e.g. https://github.com/w3c/trusted-types/issues/359), I'm not sure if those sinks could be guarded only by hooking on DOM attributes, it might require a different approach like disabling XSLT altogether.

Event handler attributes have an integration via HTML, and attributes supporting javascript: URLs are also handled in CSP integration. 

There might be xlink:href missing maybe for svg:script ? It's hard to say, pending https://github.com/w3c/svgwg/issues/196 in SVG, https://svgwg.org/svg2-draft/interact.html#ScriptElement is quite vague, but I guess we could add it here too?

Anything else I missed?

spec/index.bs
koto: > The check for the HTML namespace here is wrong.

Should I just check for null here? I have trouble reverse engineering what a null namespace value means vs HTML namespace in DOM. 
 
> The case-insensitive match seems wrong. Especially if as you say this is IDL-backed in some implementations. Can you point to a test case that exercises that code?

Do you mean wrong or unnecessary? Essentially we need to be able to check when `srCdoC` is the attribute name. Do you mean that can never happen in HTML ns?

https://wpt.fyi/results/trusted-types/Element-setAttribute.html?label=experimental&label=master&aligned (@mbrodesser-Igalia we need to add a mixed-case attribute name test case there).

Essentially I want to do the check for specific html+attr combinations. Assuming I want to future-proof it to add e.g. svg:script+xlink:href, what's the easiest language to get there?

spec/index.bs
annevk: Things I wasn't sure about:

1. frame src / iframe src
2. a href / area href / form action / * form action

I think 1 is out-of-scope and 2 is handled by CSP it sounds like.

The current list doesn't handle the SVG script element at all, so that needs to be added, yes. (It's not an HTMLScriptElement. In Chromium/WebKit they share a common class, but that's not a thing in specifications.)

koto: Yes, all of those sinks are handled by CSP blocking navigation to `javascript:`, and we don't control iframe/src beyond that.

> SVG
ACK.

spec/index.bs
annevk: An attribute in the HTML namespace would not be recognized as having any meaning. You can test using `setAttributeNS()`. So yeah, just check for null.

Similarly, if using `setAttributeNS()` you set an attribute with no namespace and a local name of `SRCDOC`, it also would not work. Perhaps you still want to block it because of re-parsing concerns? Note that you cannot test this using `setAttribute()`. That lowercases for HTML elements.

koto: Got it, thanks. Done. I found that we actually already have a mixed-case attr test in https://github.com/web-platform-tests/wpt/blob/501b929d04/trusted-types/block-string-assignment-to-Element-setAttribute.html#L63, but filed https://github.com/w3c/trusted-types/issues/425 for other cases.

annevk: As I pointed out that's not a test, right? setAttribute lowercases for HTML elements so it's kind of a useless entry point for testing this.

spec/index.bs
```diff
@@ -1113,6 +1101,39 @@ Given an {{HTMLScriptElement}} (|script|), this algorithm performs the following
...
+
+1. If |expectedType| is null, then:
+    1. If |newValue| is a string, return |newValue|.
+    1. <a>Assert</a>: |newValue| is {{TrustedHTML}} or {{TrustedScript}} or {{TrustedScriptURL}}.
+    1. Return |value|'s `[[Data]]` internal slot value.
+1. Return the result of executing [$Get Trusted Type compliant string$] with the following arguments:
+     * |expectedType|
+     * |newValue| as |input|
+     * |element|'s <a>node document</a>'s <a>relevant global object</a> as |global|
+     * the result of [=concatenating=] « |element|'s [=Element/local name=], |attribute|'s [=Attr/local name=] » with `"."` as a |separator| as |sink|
```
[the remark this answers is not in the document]
koto: Yes, we're lucky. In general the sink name is just some approximation of where the violation is occurring. Since it's present in the violation report samples, which are trimmed, we wanted to come up with a short, close-enough identifier. We can add an assert here for SVG or null namespace of the attribute, nothing else comes to mind.

spec/index.bs
koto: Agreed, #425 proposes a `setAttributeNS` as an entry point.

spec/index.bs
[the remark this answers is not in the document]
koto: Both good and bad news with the newest commit.

I've added sink name to the table and in fact, following https://github.com/w3c/trusted-types/issues/260#issuecomment-594737034m we were already using a longer form of the sink name in the implementation.

I've also verified that a `svg:script` executes code with `href` attribute from both null and xlink namespaces, so we do have a conflict in sink name already. Like explained before, I don't think it's a practical problem, because I can't find a reason why authors would treat (e.g. sanitize) both attributes values differently, branching on namespace.

mbrodesser-Igalia (commented):
Couldn't add an inline comment for https://github.com/w3c/trusted-types/pull/418/files#diff-40cc3a1ba233cc3ca7b6d5873260da9676f6ae20bb897b62f7871c80d0bda4e9R1060-L1059, hence here: don't those lines imply

```js
someIframe.setAttribute("srcdoc", "someHtml");
```
would set the srcdoc's value to "someHtml", because `someIframe`'s global differs from the embedding document's global?

Asking because of https://github.com/w3c/trusted-types/issues/425#issue-2100418157.

CC @koto

koto:
> don't those lines imply
> 
> ```js
> someIframe.setAttribute("srcdoc", "someHtml");
> ```
> 
> would set the srcdoc's value to "someHtml", because `someIframe`'s global differs from the embedding document's global?

Isn't iframe _element_ global the same? Only its [content window](https://html.spec.whatwg.org/multipage/document-sequences.html#content-window) can have a different global.

spec/index.bs
mbrodesser-Igalia: > Do we want this or the element's node document's global? Is there test coverage for this? (Gecko mutates the global when moving elements across documents which is against the specification.) I think in another case we decided to use the element's node document for cases like this, not because of Gecko, but because it would be more likely what is expected.
> 
> cc @domenic

@annevk Expected by or for what?

annevk: By web developers.

spec/index.bs
```diff
@@ -1113,6 +1113,41 @@ Given an {{HTMLScriptElement}} (|script|), this algorithm performs the following
...
+      <thead>
+        <tr><th>Element<th>Attribute namespace<th>Attribute local name<th>TrustedType<th>Sink
+      <tbody>
+        <tr><td>{{HTMLIFrameElement}}<td>null<td>"srcdoc"<td>{{TrustedHTML}}<td>"HTMLIFrameElement srcdoc"
+        <tr><td>{{HTMLEmbedElement}}<td>null<td>"src"<td>{{TrustedScriptURL}}<td>"HTMLEmbedElement src"
+        <tr><td>{{HTMLScriptElement}}<td>null<td>"src"<td>{{TrustedScriptURL}}<td>"HTMLScriptElement src"
+        <tr><td>{{HTMLObjectElement}}<td>null<td>"data"<td>{{TrustedScriptURL}}<td>"HTMLObjectElement data"
+        <tr><td>{{HTMLObjectElement}}<td>null<td>"codebase"<td>{{TrustedScriptURL}}<td>"HTMLObjectElement codebase"
+        <tr><td>{{SVGScriptElement}}<td>null<td>"href"<td>{{TrustedScriptURL}}<td>"SVGScriptElement href"
+        <tr><td>{{SVGScriptElement}}<td><a>XLink namespace</a><td>"href"<td>{{TrustedScriptURL}}<td>"SVGScriptElement href"
```
mbrodesser-Igalia, line 1134: Given https://searchfox.org/mozilla-central/rev/76cb3efe3b19e649bf675bb6ec5d4af8109b9771/testing/web-platform/tests/trusted-types/block-string-assignment-to-attribute-via-attribute-node.html#15,36 and https://jsfiddle.net/3vef62r9/1/ `onclick` should be handled here too.

CC @koto

koto: These should be handled in https://w3c.github.io/trusted-types/dist/spec/#enforcement-in-event-handler-content-attributes instead (via HTML integration).

mbrodesser-Igalia: Perhaps other (all?) event handlers too.

koto: Event handlers are intentionally not in this list, they are controlled via HTML integration. https://w3c.github.io/trusted-types/dist/spec/#enforcement-in-event-handler-content-attributes

spec/index.bs
mbrodesser-Igalia: > An attribute in the HTML namespace would not be recognized as having any meaning. You can test using `setAttributeNS()`. So yeah, just check for null.

Indeed.

That's very confusing though, see https://jsfiddle.net/zao4yhc8/3/.

CC @annevk

PR #420, opened 2024-01-24
Extract Get Trusted Type policy value algorithm

### LINKED ISSUE #414: CreateXXX as specced can return null values but not marked as such in IDL
https://w3c.github.io/trusted-types/dist/spec/#create-a-trusted-type-algorithm - returns null or undefined if that's what the default policies functions return and the default policy is being invoked.

But CreateXXX functions on TrustedTypePolicy are marked as returning TrustedHTML without any optionality.

Chromium doesn't ever seem to return null in this case either so I'm slightly confused.

lukewarlow:
This fixes an error with default policy handling

Closes #414


koto (approved):
🥇

koto:
Very elegant, merged, thanks!

PR #445, opened 2024-02-20
Update dynamic code compilation support

lukewarlow:
Changes to go along with https://github.com/tc39/proposal-dynamic-code-brand-checks/pull/10

***
<a href="https://pr-preview.s3.amazonaws.com/lukewarlow/trusted-types/pull/445.html" title="Last updated on Feb 22, 2024, 2:16 PM UTC (79dfc9b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trusted-types/445/4e858f8...lukewarlow:79dfc9b.html" title="Last updated on Feb 22, 2024, 2:16 PM UTC (79dfc9b)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1728,8 +1728,7 @@ The Trusted Types portion of this algorithm uses |calleeRealm| and its CSP setti
 </pre>
 </div>
 
-Given a [[ECMA-262#realm|realm]] (|calleeRealm|), a string
-(|source|) <ins>, a boolean |wasCodeLike| and a string |compilationSink|</ins>, this algorithm returns <del>normally</del><ins>the
+Given a [[ECMA-262#realm|realm]] (|calleeRealm|), a list of strings (|parameterStrings|), a string (|bodyString|), <ins> a string (|source|), an enum (|compilationType|), and a boolean |wasCodeLike|</ins>, this algorithm returns <del>normally</del><ins>the
 source string to compile</ins> if compilation is allowed, and
 throws an "`EvalError`" if not:
 
@@ -1737,6 +1736,7 @@ throws an "`EvalError`" if not:
     the {{TrustedScript}} interface, with its `[[Data]]` internal slot value
     set to |source|. Otherwise, let |sourceToValidate| be |source|.</ins>
 
+1. <ins>Let |compilationSink| be `"Function"` if |compilationType| is `*FUNCTION*`, otherwise `"Eval"`.</ins>
 1. <ins>Let |sourceString| be the result of executing the
     [$Get Trusted Type compliant string$] algorithm, with:
     *   |calleeRealm| as |global|,


spec/index.bs
```diff
@@ -1728,15 +1728,15 @@ The Trusted Types portion of this algorithm uses |calleeRealm| and its CSP setti
 </pre>
 </div>
 
-Given a [[ECMA-262#realm|realm]] (|calleeRealm|), a string
-(|source|) <ins>, a boolean |wasCodeLike| and a string |compilationSink|</ins>, this algorithm returns <del>normally</del><ins>the
+Given a [[ECMA-262#realm|realm]] (|calleeRealm|), a list of strings (|parameterStrings|), a string (|bodyString|), <ins> a string (|source|), an enum (|compilationType|), and a boolean |wasCodeLike|</ins>, this algorithm returns <del>normally</del><ins>the
```
lukewarlow, line 1731: We potentially could pass bodyString and parameterStrings as extra params to the default policy?

This could give policy authors a nicer format of data for them to work with, without needing to first parse the function string?

This wouldn't help them with manipulating it back into a stringified function though?

I feel like we could make the ergonomics of this better for the Function constructor though.

lukewarlow: Having said that they can just do `/\(function anonymous\((.*,+.*)\\n\) {\\n(.*)\\n}\)/` to capture the csv of params and the body string process them and then recreate the string to return?

koto: I kinda like the ergonomics of separating the function arguments. There is still a problem as the default policy can't tell which one(s) of the arguments/body wasn't a TT initially, but it might be a bit too involved to check all of them separately in a callback (and we'd probably need a `TrustedFunctionArgument` at least). 

As is, the PR doesn't use the passed arguments. I'm happy to merge, but this is aligned with https://specs.lukewarlow.dev/dynamic-code-brand-checks/; do you know how likely is that shape to be merged to ECMAScript? Not separating arguments is an easier integration for sure.

cc @otherdaniel too.

lukewarlow: So the separate args for body and parameter are already merged, this would include the fully compiled string.

The separate body and parameter are needed for other host functionality such as extensions to CSP (hashing).

Ideally we wouldn't need to additionally pass the fully compiled string but the default policy needs it to be web compatible. So we'd either need to separately compile it ourselves or pass it through.

I think it could be useful to pass through the separate values to the default policy additionally (this would be something that's different from Chrome's current implementation).

To clarify do you agree we should also pass them through as arguments to the default policy @koto I can update this PR to do so quite easily

lukewarlow: As for how likely that proposal is to get into ECMAScript idk. The web compatability impacting change (reording of stringifier calls) has already occured outside of this proposal so that's one contentious bit down.

As far as I see it the HostDefinedCodeLike slot is the thing most likely to stop it. But that's an academic discussion as we'd be using a slot in web land even if not in ecma land...

lukewarlow: @koto before merging this it would be good to get the proposal update PR merged too that way they're both inline with each other

koto: OK let's do it. Since for now the PR doesn't change the default policy behavior, there's no compatibility risk. I'd like @otherdaniel to comment on the Function constructor + the changes to the default policy arguments, but we have https://issues.chromium.org/issues/40133092, so there's probably no compatibility risk whatsoever.

koto (approved):

spec/index.bs
lukewarlow: I thought so too but that's just that TrustedScript isn't "trusted" the strings still go through the default policy which is the compat risk (if we change arguments).

lukewarlow: https://github.com/w3c/trusted-types/issues/458 follow up issue

PR #450, opened 2024-02-21
Extract `Does sink type require trusted types?` to its own algorithm

lukewarlow:
In it's current form it was rather hard to read through, this extracts it to the CSP section and fixes some grammar issues.


lukewarlow:
cc @mbrodesser-Igalia does this seem like a reasonable change? Just something I came across when cleaning up a PR to webkit to implement it. I was trying to follow the spec as close as possible and that line tripped me up massively.

koto:
Not sure what's with the preview, but there's not even a branch existing in here that would have a change in that directive. 

Let's monitor, but merging the code in the meantime.

koto:
`master` is way behind, it had other changes that didn't surface in the diff.

lukewarlow:
@koto https://github.com/w3c/trusted-types/actions/runs/8007592166/job/21872070992 it seems the runner has failed and not published the update.

mbrodesser-Igalia:
> Not sure what's with the preview, but there's not even a branch existing in here that would have a change in that directive.
> 
> Let's monitor, but merging the code in the meantime.

Let's please fix this, see https://github.com/w3c/trusted-types/issues/455.

Issue #423, opened 2024-01-25
getAttributeType() needs a rewrite


annevk:
It should reuse the logic from #418.

lukewarlow:
Based on #384 I'm removing the proposed-removal label

PR #457, opened 2024-02-26
Rewrite metadata functions

### LINKED ISSUE #456: `getPropertyType()` needs a rewrite?
Like https://github.com/w3c/trusted-types/issues/423 but for `getPropertyType()`.

Neither `getPropertyType()` nor `getAttributeType()` are actually implemented as specced they both in reality use a hardcoded "map" it would be better to spec both of these as actually implemented. This will help future implementators but also help drive the point that if a new sink is added to the spec it needs adding to these maps too.

lukewarlow:
- getAttributeType and getPropertyType now use lookup tables.

Fixes #456 and #423, and #474

***
<a href="https://pr-preview.s3.amazonaws.com/lukewarlow/trusted-types/pull/457.html" title="Last updated on Jun 12, 2024, 12:35 PM UTC (21eb4b4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trusted-types/457/af8680e...lukewarlow:21eb4b4.html" title="Last updated on Jun 12, 2024, 12:35 PM UTC (21eb4b4)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -670,14 +670,24 @@ Its value is initially « ».
     1. Set |localName| to |tagName| in [=ASCII lowercase=].
     1. If |elementNs| is an empty string, set |elementNs| to [=HTML namespace=].
     1. Let |interface| be the [=element interface=] for |localName| and |elementNs|.
-    1. If |interface| has an IDL <a spec="webidl">attribute</a> member which identifier is |attribute|, and
-        {{StringContext}} IDL extended attribute appears on that attribute, return
-        stringified {{StringContext}}'s identifier and abort further steps.
+    1. Let |expectedType| be null.
+    1. Find the row in the following table, where the first column is "*" or |interface|'s name, and |property| is in the second column.
+        If a matching row is found, set |expectedType| to the value of the third column.
 
-        Note: This also takes into account all members of [=interface mixins=] that
-        |interface| [=includes=].
-
-    1. Return null.
+        <table>
+          <thead>
+            <tr><th>Element<th>Property name<th>TrustedType
+          <tbody>
+            <tr><td>{{HTMLIFrameElement}}<td>"srcdoc"<td>{{TrustedHTML}}
+            <tr><td>{{HTMLScriptElement}}<td>"innerText"<td>{{TrustedScript}}
+            <tr><td>{{HTMLScriptElement}}<td>"src"<td>{{TrustedScriptURL}}
+            <tr><td>{{HTMLScriptElement}}<td>"text"<td>{{TrustedScript}}
+            <tr><td>{{HTMLScriptElement}}<td>"textContent"<td>{{TrustedScript}}
+            <tr><td>"*"<td>"innerHTML"<td>{{TrustedHTML}}
+            <tr><td>"*"<td>"outerHTML"<td>{{TrustedHTML}}
+          </tbody>
+        </table>
+    1. Return |expectedType|.
 
     <div class="example" id="get-property-type-example">
     <xmp highlight="js">
@@ -699,12 +709,13 @@ Its value is initially « ».
     1. If |elementNs| is an empty string, set |elementNs| to [=HTML namespace=].
     1. If |attrNs| is an empty string, set |attrNs| to null.
     1. Let |interface| be the [=element interface=] for |localName| and |elementNs|.
-    1. If |interface| does not have an IDL <a spec="webidl">attribute</a> that [=reflects=] a content attribute with
-        |localName| local name and |attrNs| namespace,
-        return undefined and abort further steps. Otherwise, let |idlAttribute| be that IDL <a spec="webidl">attribute</a>.
-    1. If {{StringContext}} IDL extended attribute appears on |idlAttribute|, return
-        stringified {{StringContext}}'s identifier and abort further steps.
-    1. Return null.
+    1. Let |expectedType| be null.
+    1. Set |attributeData| to the result of [$Get Trusted Type data for attribute$] algorithm, with the following arguments:
+        * |interface| as |element|
+        * |attribute|
+        * |attrNs|
+    1. If |attributeData| is not null, then set |expectedType| to the value of the third member of |attributeData|.
+    1. Return |expectedType|.
 
     <div class="example" id="get-attribute-type-example">
     <xmp highlight="js">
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -672,14 +672,27 @@ initially empty.
...
-        |interface| [=includes=].
-
-    1. Return null.
+    1. Let |expectedType| be null.
+    1. Find the row in the following table, where the first column is "*" or |interface|'s name, and |property| is in the second column.
+        If a matching row is found, set |expectedType| to the value of the third column.
+
+        <table>
+          <thead>
+            <tr><th>Element<th>Property name<th>TrustedType
```
lukewarlow, line 681: I feel like this table needs a baseVal property for the SVGAnimatedString type somehow, not sure on the specifics of that SVG integration (I also am not sure that baseVal is accounted for in Chromium if it is required in this list).

spec/index.bs
```diff
@@ -701,12 +714,24 @@ initially empty.
...
-    1. If {{StringContext}} IDL extended attribute appears on |idlAttribute|, return
-        stringified {{StringContext}}'s identifier and abort further steps.
-    1. Return null.
+    1. Let |expectedType| be null.
+    1. Find the row in the following table, where |interface|'s name is in the first column, and |attribute|'s <a for="Attr">namespace</a> and <a for="Attr">local name</a> match the values in the second and third column, respectively.
+        If a matching row is found, set |expectedType| to the value of the fourth.
+
+        <table>
+          <thead>
+            <tr><th>Element<th>Attribute namespace<th>Attribute local name<th>TrustedType
```
lukewarlow, line 723: This is duplicated from Get trusted type compliant attribute value, it should probably be shared somehow (the other function needs a sink name too else I'd make an alogrithm that returns the value).

Could I make a shared algorithm that returns the row somehow?

spec/index.bs
```diff
@@ -701,12 +714,24 @@ initially empty.
...
+          <tbody>
+            <tr><td>{{HTMLIFrameElement}}<td>null<td>"srcdoc"<td>{{TrustedHTML}}
+            <tr><td>{{HTMLEmbedElement}}<td>null<td>"src"<td>{{TrustedScriptURL}}
+            <tr><td>{{HTMLScriptElement}}<td>null<td>"src"<td>{{TrustedScriptURL}}
+            <tr><td>{{HTMLObjectElement}}<td>null<td>"data"<td>{{TrustedScriptURL}}
+            <tr><td>{{HTMLObjectElement}}<td>null<td>"codebase"<td>{{TrustedScriptURL}}
+            <tr><td>{{SVGScriptElement}}<td>null<td>"href"<td>{{TrustedScriptURL}}
+            <tr><td>{{SVGScriptElement}}<td><a>XLink namespace</a><td>"href"<td>{{TrustedScriptURL}}
```
lukewarlow, line 731: This table has no handling for event handlers. Now that as a table entry would be far too long and brittle.

I'm thinking of before this doing if attribute starts with "on" just return "TrustedScript"?

spec/index.bs
```diff
@@ -672,14 +672,27 @@ initially empty.
...
+            <tr><td>{{HTMLEmbedElement}}<td>"src"<td>{{TrustedScriptURL}}
+            <tr><td>{{HTMLIFrameElement}}<td>"srcdoc"<td>{{TrustedHTML}}
+            <tr><td>{{HTMLObjectElement}}<td>"codebase"<td>{{TrustedScriptURL}}
+            <tr><td>{{HTMLObjectElement}}<td>"data"<td>{{TrustedScriptURL}}
+            <tr><td>{{HTMLScriptElement}}<td>"innerText"<td>{{TrustedScript}}
+            <tr><td>{{HTMLScriptElement}}<td>"src"<td>{{TrustedScriptURL}}
+            <tr><td>{{HTMLScriptElement}}<td>"text"<td>{{TrustedScript}}
+            <tr><td>{{HTMLScriptElement}}<td>"textContent"<td>{{TrustedScript}}
+            <tr><td>"*"<td>"innerHTML"<td>{{TrustedHTML}}
+            <tr><td>"*"<td>"outerHTML"<td>{{TrustedHTML}}
```
lukewarlow, line 692: This table has no handling for event handlers. Now that as a table entry would be far too long and brittle.

I'm thinking of before this doing if property starts with "on" just return "TrustedScript"?

lukewarlow:
@mbrodesser-Igalia do you think you'd be able to give this a review?

spec/index.bs
```diff
@@ -671,15 +671,29 @@ initially empty.
 
     1. Set |localName| to |tagName| in [=ASCII lowercase=].
     1. If |elementNs| is an empty string, set |elementNs| to [=HTML namespace=].
+    1. If |property| starts with "on", then return {{TrustedScript}}.
```
mbrodesser-Igalia, line 674: Had to look it up, so perhaps worth linking from a note to https://html.spec.whatwg.org/#event-handler-name?

E.g.
```
 <p class="note" role="note"><span class="marker">Note:</span> ...
```

lukewarlow: Ah good point

mbrodesser-Igalia:
> > @mbrodesser-Igalia do you think you'd be able to give this a review?
> 
> If this could wait until I actually start implementing it in Gecko, that might be most effective.

Thorough reviews are time-consuming.

spec/index.bs
```diff
@@ -671,15 +671,31 @@ initially empty.
...
```
lukewarlow, line 674: Need to remove this step from getPropertyType because it's only attributes that can be event handlers

spec/index.bs
```diff
@@ -698,15 +714,19 @@ initially empty.
 
     1. Set |localName| to |tagName| in [=ASCII lowercase=].
     1. Set |attribute| to |attribute| in [=ASCII lowercase=].
+    1. If |attribute| starts with "on", then return {{TrustedScript}}.
```
lukewarlow, line 717: There's probably a better way of doing this? The question is do we care? If someone does `getAttributeType('div', 'one')` and it says TrustedScript is that an issue?

Possibly could do this via https://html.spec.whatwg.org/#event-handler-content-attributes some lookup based on this concept?

lukewarlow: Fwiw Chrome currently does something more safisticated than just looking for the on prefix.

koto: Correct, we should only be blocking on actual event handler attributes.

lukewarlow: Fixed this

lukewarlow: This also addresses #474

spec/index.bs
```diff
@@ -1100,7 +1101,28 @@ To <dfn abstract-op export>get Trusted Types-compliant attribute value</dfn> on
 
    If the algorithm threw an error, rethrow the error.
 
-Issue: This algorithm should account for event handler attributes. See https://github.com/w3c/trusted-types/issues/474
+## Get Trusted Type data for attribute ## {#get-trusted-type-data-for-attribute}
+To <dfn abstract-op>Get Trusted Type data for attribute</dfn> given |element|, |attribute|, |attributeNs|, perform the following steps:
+
+1. Let |data| be null.
+1. If |attributeNs| is null, and |attribute| is the name of an [=event handler content attribute=], then:
```
annevk, line 1108: I'm not really sure how you envision this lookup to work. There are many event handler content attributes that are not applicable to elements. And some only apply to a specific element.

lukewarlow: > There are many event handler content attributes

I don't fully understand what you mean by this? What event handler content attribute doesn't apply to an element?
Do you mean legacy ones no longer applicable or something like that?

> And some only apply to a specific element.

So for the purposes of trusted types I don't think we mind being overly restrictive here, it seems fine to consider them independent of the element (this is the case in Chromium's implementation)?

For context the plan in WebKit is to use `HTMLElement::eventNameForEventHandlerAttribute` with a null check.

annevk: `onreadystatechange`?

I'm not sure we have a specification concept for `eventNameForEventHandlerAttribute`.

lukewarlow: Chrome enforces the TT on that attribute for a div element for example, so that aspect is fine.

lukewarlow: If I understand correctly the issue would be that you require a trusted object to set an inline event listener that is "safe" for a given element. Given inline event listeners are bad practice and a strict csp would already block them from actually running I don't think blocking "harmless" but otherwise useless (like are people actually going to be trying to set onreadystatechange on an arbitrary element?) attributes from being set from a string is fine?

lukewarlow: onloadend also requires a trusted script to be set on div

annevk: What about `onblah`?

My worry here is about the hook not being defined in sufficient detail for an interoperable implementation. Implementing "is an event handler content attribute" requires knowing about all event handler attributes in existence, not just those defined in HTML.

lukewarlow: `onblah` is allowed. I *think* it's using IDL to generate the list from `attribute EventHandler`s `onrepeat` for example which as far as I can tell is only SVG related is being blocked.

lukewarlow: Given we can't use the on* pattern what is the next best steps here?

lukewarlow: Both Chrome and WebKit have a way to look up EventHandlers from IDL which gives us the coverage we need to ensure stuff is protected. Idk if that corresponds to the current spec text, or how we would spec that?

The key bit is that all sinks for a given browser are included in that browser, which the spec as written I think covers?

annevk: I think we should have an actual list of event handler attributes we care about (those on elements), which would be a subset of all event handler attributes. That's also what the Sanitizer API would need.

I don't think we want to do something with event handler attributes that only have meaning on a non-element object.

spec/index.bs
lukewarlow: https://github.com/w3c/trusted-types/issues/521

spec/index.bs
lukewarlow: Have made a new issue to track this and added an inline issue in the spec referencing it. https://github.com/w3c/trusted-types/issues/520

lukewarlow:
For the sake of keeping the spec moving I've made two new issues to track the remaining questions here and am going to get this merged in so we can remove the last StringContext attribute mentions. The whole spec needs an editorial run through and any issues here can be addressed as part of that or smaller follow ups.

PR #481, opened 2024-03-14
Update handling of timer functions

### LINKED ISSUE #480: HTML timers as specced won't work
As currently specced setTimeout and setInterval won't always work as expected. We used the String context attribute which works fine, but this means that a plain string is passed to HostEnsureCanCompileStrings which TT will reject (or pass into the default policy).

See https://w3c.github.io/trusted-types/dist/spec/#enforcement-in-timer-functions

Also See https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#timer-initialisation-steps for full steps.

cc @koto @otherdaniel @mbrodesser-Igalia

lukewarlow:
Fixes #480 

This effectively follows the Chromium implementation, we use the StringContext attribute to do the enforcement for setTimeout and setInterval and then update the `HostEnsureCanCompileStrings` call (which now actually calls the CSP function directly) accordingly.

We then update `EnsureCSPDoesNotBlockStringCompilation` to only do TT validation for eval and Function, not timers.

***
<a href="https://pr-preview.s3.amazonaws.com/lukewarlow/trusted-types/pull/481.html" title="Last updated on Mar 26, 2024, 4:45 PM UTC (2c0de9f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trusted-types/481/1eee73a...lukewarlow:2c0de9f.html" title="Last updated on Mar 26, 2024, 4:45 PM UTC (2c0de9f)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1248,14 +1248,14 @@ partial interface mixin WindowOrWorkerGlobalScope {
 };
 </pre>
 
-To the [[HTML5#timer-initialisation-steps|timer initialization steps algorithm]],
-change step 8.3.2 as follows:
-
-1.  Perform <del>HostEnsureCanCompileStrings</del><ins>EnsureCSPDoesNotBlockStringCompilation</ins>(<var>realm</var>, « », <var>handler</var>,<del> false</del><ins>, <var>handler</var>, ~timer~, « », <var>handler</var></ins>). If this throws an exception, catch it, report the exception, and abort these steps.
-
 Note: This makes sure that a {{TrustedScript}} is passed to timer
 functions in place of a string when Trusted Types are enforced, but
 also unconditionally accepts any {{Function}} object.
+
+To the [[HTML5#timer-initialisation-steps|timer initialization steps algorithm]],
+change step 8.4.3 as follows:
+
+1.  Perform <del>HostEnsureCanCompileStrings</del><ins>EnsureCSPDoesNotBlockStringCompilation</ins>(<var>realm</var>, « », <var>handler</var>,<del> false</del><ins>, <var>handler</var>, ~timer~, « », <var>handler</var></ins>). If this throws an exception, catch it, report the exception, and abort these steps.
 
 ### Enforcement in event handler content attributes ### {#enforcement-in-event-handler-content-attributes}
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1249,17 +1249,9 @@ partial interface mixin WindowOrWorkerGlobalScope {
 </pre>
 
 To the [[HTML5#timer-initialisation-steps|timer initialization steps algorithm]],
-add this step between 7.1 and 7.2:
+change step 8.3.2 as follows:
 
-1.  If the first operation argument is not a {{Function}}, set the first operation argument to the result of executing
-    the [$Get Trusted Type compliant string$] algorithm, with
-    *   |global| set to the [=this=] value's [=relevant global object=].
-    *   |input| set to the first method argument, and
-    *   |expectedType| set to {{TrustedScript}}.
-    *   |sink| set to `Window setInterval` if <var ignore>repeat</var> is true, `Window setTimeout` otherwise.
-    *   |sinkGroup| set to `'script'`.
-
-        Note: This matches the logic that the extended attribute would apply.
+1.  Perform HostEnsureCanCompileStrings(<var>realm</var>, « », <var>handler</var>,<del> false</del><ins>, <var>handler</var>, ~timer~, « », <var>handler</var></ins>). If this throws an exception, catch it, report the exception, and abort these steps.
 
 Note: This makes sure that a {{TrustedScript}} is passed to timer
 functions in place of a string when Trusted Types are enforced, but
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1249,17 +1249,9 @@ partial interface mixin WindowOrWorkerGlobalScope {
...
 
-1.  If the first operation argument is not a {{Function}}, set the first operation argument to the result of executing
-    the [$Get Trusted Type compliant string$] algorithm, with
-    *   |global| set to the [=this=] value's [=relevant global object=].
-    *   |input| set to the first method argument, and
-    *   |expectedType| set to {{TrustedScript}}.
-    *   |sink| set to `Window setInterval` if <var ignore>repeat</var> is true, `Window setTimeout` otherwise.
-    *   |sinkGroup| set to `'script'`.
-
-        Note: This matches the logic that the extended attribute would apply.
```
lukewarlow, line 1262: We use the string context here, so no need to do any of this block.

spec/index.bs
```diff
@@ -1249,17 +1249,9 @@ partial interface mixin WindowOrWorkerGlobalScope {
...
+1.  Perform HostEnsureCanCompileStrings(<var>realm</var>, « », <var>handler</var>,<del> false</del><ins>, <var>handler</var>, ~timer~, « », <var>handler</var></ins>). If this throws an exception, catch it, report the exception, and abort these steps.
```
lukewarlow, line 1254: The `~timer~` enum value here might need adding to https://github.com/tc39/ecma262/pull/3294 even though it'll only be used by HTML and TT? @nicolo-ribaudo am I correct in saying that?

spec/index.bs
```diff
@@ -1249,17 +1249,9 @@ partial interface mixin WindowOrWorkerGlobalScope {
 </pre>
```
mbrodesser-Igalia, line 1249: Line 1243: not part of this patch, but it seems necessary to understand and document it. Why is `or Function` acceptable here? This seems to be a hole in TT.

lukewarlow: Because a function isn't dangerous?
It's just so you can do like setTimeout(() =>  delayedThing(), 500);

spec/index.bs
mbrodesser-Igalia, line 1254: Why isn't `HostEnsureCanCompileStrings` patched first? It seems that'd be cleaner. https://html.spec.whatwg.org/#timer-initialisation-steps takes different arguments than the ones passed here.

spec/index.bs
mbrodesser-Igalia: > Because a function isn't dangerous? It's just so you can do like setTimeout(() => delayedThing(), 500);

What if `x` is user-controlled:

```
let f = () => { console.log(x ? "do good" : "do bad" )}
```

mbrodesser-Igalia: Right, TT can't prevent that and the problem is unrelated to `setTimeout`. Feel free to resolve this conversation.

spec/index.bs
lukewarlow: HostEnsureCanCompileStrings will be updated accordingly, don't think it's worth adding it in this PR. https://w3c.github.io/trusted-types/dist/spec/#csp-eval already updates EnsureCSPDoesNotBlockStringCompilation so it's really just the HTML HostEnsureCanCompileStrings block that isn't handled.

Once the corresponding tc39 change gets merged it'll make the layering of all these changes much easier.

lukewarlow: https://github.com/w3c/trusted-types/pull/490 - made a PR to update the host function in HTML land.

lukewarlow: The answer is no the ECMAScript PR doesn't need this.

otherdaniel (commented):
Thank you. Generally, this looks exactly right to me!

There's one editorial nitpick with the note.
The other issue is about the definition of `|compilationType|`, where I might well be overlooking something. Would be good to get that resolved.

spec/index.bs
```diff
@@ -1668,32 +1660,35 @@ The Trusted Types portion of this algorithm uses |calleeRealm| and its CSP setti
...
-    *   |calleeRealm| as |global|,
-    *   |sourceToValidate| as |input|,
-    *   |compilationSink| as |sink|,
-    *   `'script'` as |sinkGroup|,
-    *   {{TrustedScript}} as |expectedType|.</ins>
-
-1.  <ins>If the algorithm throws an error, throw an {{EvalError}}.</ins>
-
-1.  <ins>If |sourceString| is not equal to |source|, throw an {{EvalError}}.</ins>
+1.  <ins>If |compilationType| is `*TIMER*`, then:</ins>
```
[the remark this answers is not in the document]
lukewarlow: > Given a [[ECMA-262#realm|realm]] (|calleeRealm|), a list of strings (|parameterStrings|), a string (|bodyString|), <ins> a string (|source|), **an enum (|compilationType|),** a list of ECMAScript language values (|parameterArgs|), and an ECMAScript language value (|bodyArg|), this algorithm returns normally if compilation is allowed, and

It's defined here, it's can have a value of eval, function, or timer.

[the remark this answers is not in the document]
lukewarlow: We could do a boolean Param tbf but I think given the large number of params and the fact that the csp function is currently a perfect match to the ecma host hook, the benefits don't outweigh the downsides.

spec/index.bs
```diff
@@ -1249,17 +1249,9 @@ partial interface mixin WindowOrWorkerGlobalScope {
 </pre>
 
 To the [[HTML5#timer-initialisation-steps|timer initialization steps algorithm]],
-add this step between 7.1 and 7.2:
+change step 8.3.2 as follows:
```
mbrodesser-Igalia, line 1252: In https://html.spec.whatwg.org/#timer-initialisation-steps. that's now 8.3.4. Maybe that changes again?
Sorry for the nit-picking but from experience, specs sometimes stay unintendedly in 'temporary' states. Therefore each patch should not rely on other patches to come.

spec/index.bs
```diff
@@ -1249,17 +1249,9 @@ partial interface mixin WindowOrWorkerGlobalScope {
...
-1.  If the first operation argument is not a {{Function}}, set the first operation argument to the result of executing
-    the [$Get Trusted Type compliant string$] algorithm, with
-    *   |global| set to the [=this=] value's [=relevant global object=].
-    *   |input| set to the first method argument, and
-    *   |expectedType| set to {{TrustedScript}}.
-    *   |sink| set to `Window setInterval` if <var ignore>repeat</var> is true, `Window setTimeout` otherwise.
-    *   |sinkGroup| set to `'script'`.
-
-        Note: This matches the logic that the extended attribute would apply.
+1.  Perform <del>HostEnsureCanCompileStrings</del><ins>EnsureCSPDoesNotBlockStringCompilation</ins>(<var>realm</var>, « », <var>handler</var>,<del> false</del><ins>, <var>handler</var>, ~timer~, « », <var>handler</var></ins>). If this throws an exception, catch it, report the exception, and abort these steps.
```
mbrodesser-Igalia, line 1254: This should link to https://w3c.github.io/webappsec-csp/#can-compile-strings.

mbrodesser-Igalia: Oh, that's modified in this patch too.

TBH it would be much clearer if the spec changes were upstreamed and merged in a way where the upstream changes don't break upstream. I.e., couldn't https://pr-preview.s3.amazonaws.com/w3c/trusted-types/481/1eee73a...lukewarlow:2c0de9f.html#csp-eval get upstreamed first?

lukewarlow: Yeah I'm definitely going to upstream things as soon as possible. It's all waiting on the tc39 lynchpin.

lukewarlow:
Based on @otherdaniel review I'm going to go ahead and merge this

Issue #483, opened 2024-03-18
SVGScriptElement needs TT protection too


lukewarlow:
I know the script enforcement is currently in the HTML integrations section but we need to update it (possibly move the whole section?) to include coverage of the SVGScriptElement. At the very least it needs the new `[[ScriptText]]` slot to be handled by the  parser script enforcement (see #252 )

See also #437

annevk:
Reopening as #533 got reverted in #565.

Issue #437, opened 2024-02-19
Issue with script enforcement


lukewarlow:
Apologies if I'm missing something but I believe that the spec as currently written blocks any inline script elements from executing.

https://w3c.github.io/trusted-types/dist/spec/#enforcement-in-scripts

A new [[ScriptText]] slot is added and is initially null, this is only every set via the IDL OR when running "prepare the script text" algorithm.

The problem is that the "prepare the script text" algorithm will always throw unless a default policy exists. This means that the test suite for example fails to run.

Chromium I believe has an additional step that parsed script elements set their [[ScriptText]] slot when parsing has finished, this needs speccing.

mbrodesser-Igalia:
> Apologies if I'm missing something but I believe that the spec as currently written blocks any inline script elements from executing.
> 
> https://w3c.github.io/trusted-types/dist/spec/#enforcement-in-scripts
> 
> A new [[ScriptText]] slot is added and is initially null, this is only every set via the IDL OR when running "prepare the script text" algorithm.
> 

That section also mentions "Equivalent to script’s [child text content](https://dom.spec.whatwg.org/#concept-child-text-content). Initially null.". Perhaps "Equivalent to script’s [child text content](https://dom.spec.whatwg.org/#concept-child-text-content)" should be changed to a non-normative note.

> The problem is that the "prepare the script text" algorithm will always throw unless a default policy exists. This means that the test suite for example fails to run.
> 
> Chromium I believe has an additional step that parsed script elements set their [[ScriptText]] slot when parsing has finished, this needs speccing.

koto:
@otherdaniel do you remember how the parser sets the slot value? What algorithm is it a part of?

lukewarlow:
I also don't think the slot can be initially null either.

`data:text/html,<meta http-equiv="Content-Security-Policy" content="require-trusted-types-for 'script'; trusted-types default;"><div id="container"></div><script>const s = document.createElement('script'); container.appendChild(s);</script`

Else something like this would trigger a trusted types violation as it's slot is null and it's child text content is an empty string?

lukewarlow:
So I did some digging and in Chromium it's done in the FinishParsingChildren function

```c++
void HTMLScriptElement::FinishParsingChildren() {
  Element::FinishParsingChildren();

  // We normally expect the parser to finish parsing before any script gets
  // a chance to manipulate the script. However, if script parsing gets
  // deferrred (or similar; see crbug.com/1033101) then a script might get
[...]

Issue #252, opened 2019-12-13
Set slot values when called directly by the parser


koto:
Scripts created by the parser don't have their slot values filled, which formally would cause the default policy invocation in [prepare script url and text](https://w3c.github.io/webappsec-trusted-types/dist/spec/#prepare-script-url-and-text) - and might cause the rejection of the values at parsing time since [create an element for the token](https://html.spec.whatwg.org/#create-an-element-for-the-token) will append content attributes, to which we're adding validate steps.

When writing the attribute validate steps for scripts, we should accept the value and bail out if the algorithm was called from within HTML parser. We should also set the slot value in attribute change steps for `script.src` (and, later, for `iframe.srcdoc`). In fact, it's easier to move the slot setting for scripts to attribute change steps (right now it's defined at IDL level in https://w3c.github.io/webappsec-trusted-types/dist/spec/#setting-slot-values).

I'm not sure yet what to do with script bodies.

koto:
This is the Chrome implementation (only script text is affected): https://chromium-review.googlesource.com/c/chromium/src/+/2041622

koto:
See also https://bugs.chromium.org/p/chromium/issues/detail?id=1218746

PR #484, opened 2024-03-18
Update IDL for script enforcement

lukewarlow:
- Node/textContent, and Element/innerText are both now shadowed on HTMLScriptElement.

~HTMLScriptElement/textContent uses a union type rather than [StringContext] because it uses a nullable type.~

All of the properties are using union types as that seems to be the direction the spec is going to go in. But I can revert that commit and go back to using the extended attribute.

This PR is part of 1 of addressing #437 it specifically doesn't address https://github.com/w3c/trusted-types/issues/483 or https://github.com/w3c/trusted-types/issues/252

***
<a href="https://pr-preview.s3.amazonaws.com/lukewarlow/trusted-types/pull/484.html" title="Last updated on Apr 22, 2024, 4:02 PM UTC (80b9076)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trusted-types/484/74e413e...lukewarlow:80b9076.html" title="Last updated on Apr 22, 2024, 4:02 PM UTC (80b9076)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1174,7 +1174,7 @@ The {{HTMLScriptElement/innerText}} setter steps are:
 1.  Let |value| be the result of calling [$Get Trusted Type compliant string$] with
     {{TrustedScript}}, [=this=]'s [=relevant global object=], the given value, `HTMLScriptElement innerText`, and
     `script`.
-1.  Set [=script text=] value to |value|.
+1.  Set [=this=]'s [=script text=] value to |value|.
 1.  Run [=set the inner text steps=] with [=this=] and |value|.
 
 The {{HTMLScriptElement/innerText}} getter steps are:
@@ -1189,7 +1189,7 @@ empty string instead, and then do as described below:
 1.  Let |value| be the result of calling [$Get Trusted Type compliant string$] with
     {{TrustedScript}}, [=this=]'s [=relevant global object=], the given value, `HTMLScriptElement textContent`, and
     `script`.
-1.  Set [=script text=] value to |value|.
+1.  Set [=this=]'s [=script text=] value to |value|.
 1.  Run [=set text content=] with [=this=] and |value|.
 
 The {{HTMLScriptElement/textContent}} getter steps are:
@@ -1203,8 +1203,8 @@ Update the {{HTMLScriptElement/text}} setter steps algorithm as follows.
 1.  <ins>Let |value| be the result of calling [$Get Trusted Type compliant string$] with
     {{TrustedScript}}, [=this=]'s [=relevant global object=], the given value, `HTMLScriptElement text`, and
     `script`.</ins>
-1.  <ins>Set [=script text=] value to the given value.</ins>
-1.  [=String replace all=] with the given value within this <code>script</code> element.
+1.  <ins>Set [=this=]'s [=script text=] value to the given value.</ins>
+1.  [=String replace all=] with the given value within this.
 
 
 #### The {{HTMLScriptElement/src}} IDL attribute #### {#the-src-idl-attribute}
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -122,6 +122,7 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
     type: dfn; text: prepare the script element; url: prepare-the-script-element
     type: dfn; text: get the text steps; url: get-the-text-steps
     type: dfn; text: set the inner text steps; url: set-the-inner-text-steps
+    type: dfn; text: src; url: attr-script-src
 spec:DOM; urlPrefix: https://dom.spec.whatwg.org/
     type: dfn; text: get text content; url: get-text-content
     type: dfn; text: set text content; url: set-text-content
@@ -1144,16 +1145,16 @@ type policy factory]].
 
 This document modifies how {{HTMLScriptElement}} [=child text content=] can be set to allow applications to control dynamically created scripts. It does so by
 adding the {{HTMLElement/innerText}} and {{Node/textContent}} attributes directly on {{HTMLScriptElement}}. The behavior of the attributes remains the same
-as in their original counterparts, apart from additional behavior triggered by the {{StringContext}} extended attribute presence.
+as in their original counterparts, apart from the additional behavior of calling [$Get Trusted Type compliant string$].
 
 Note: Using these IDL attributes is the recommended way of dynamically setting the URL or a text of a script. Manipulating attribute nodes or text nodes directly will call a default policy on the final value when the script is prepared.
 
 <pre class="idl exclude">
 partial interface HTMLScriptElement {
- [CEReactions] attribute [LegacyNullToEmptyString] ScriptString innerText;
+ [CEReactions] attribute ([LegacyNullToEmptyString] DOMString or TrustedScript) innerText;
  [CEReactions] attribute (DOMString or TrustedScript)? textContent;
- [CEReactions] attribute ScriptURLString src;
- [CEReactions] attribute ScriptString text;
+ [CEReactions] attribute (USVString or TrustedScriptURL) src;
+ [CEReactions] attribute (DOMString or TrustedScript) text;
 };
 </pre>
 
@@ -1163,13 +1164,16 @@ This document modifies {{HTMLScriptElement}}s. Each script has:
 
 : an associated string <dfn export for="HTMLScriptElement">script text</dfn>.
 ::  A string, containing the body of the script to execute that was set
-    through a {{StringContext}} compliant sink. Equivalent to script's
+    through a compliant sink. Equivalent to script's
     [=child text content=]. Initially an empty string.
 
 #### The {{HTMLScriptElement/innerText}} IDL attribute #### {#the-innerText-idl-attribute}
 
 The {{HTMLScriptElement/innerText}} setter steps are:
 
+1.  Let |value| be the result of calling [$Get Trusted Type compliant string$] with
+    {{TrustedScript}}, [=this=]'s [=relevant global object=], the given value, `HTMLScriptElement innerText`, and
+    `script`.
 1.  Set [=script text=] value to |value|.
 1.  Run [=set the inner text steps=] with [=this=] and |value|.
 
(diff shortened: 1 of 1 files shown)


lukewarlow:
cc @mbrodesser-Igalia might be of interest to you as you start implementing more in Gecko

spec/index.bs
```diff
@@ -1186,36 +1186,94 @@ partial interface ShadowRoot {
...
+This document modifies how {{HTMLScriptElement}} [=child text content=] can be set to allow applications to control dynamically created scripts. It does so by
+modifying the {{HTMLElement/innerText}} and {{Node/textContent}} attributes. The behavior of the attributes remains the same
+as in their original counterparts, apart from additional behavior similar to that triggered by the {{StringContext}} extended attribute.
+
+Note: Using these IDL attributes is the recommended way of dynamically setting the URL or a text of a script. Manipulating attribute nodes or text nodes directly will call a default policy on the final value when the script is prepared.
+
+<pre class="idl exclude">
+partial interface Node {
+  [CEReactions] attribute (DOMString or TrustedScript)? textContent;
+};
```
annevk, line 1198: This seems extremely ungainly (same for `innerText`). Given that there's a fallback, can we just rely on the fallback and only make `.text` magic?

lukewarlow: Unfortunately we probably can't as it wouldn't be compatible with existing code.

Currently setting textContent or innerText to a trusted type would work, if we made this change it would be treated like any other string and require the default policy. Which sites might not even implement.

annevk: Hmm, maybe we could shadow these members on `script` elements, but I'm not sure we should pollute the entire node tree with this stuff.

cc @smaug----

lukewarlow: Yeah if we could find a way to only touch the script element IDL that'd be ideal.

koto: We still need to cover for the case when the setters from the `Node` are used on scripts, no? Otherwise that's a bypass.

smaug----: Not sure how this could work on script element only. One could still use the setter from Node interface, no?
@petervanderbeken

annevk: So how is setting `textNode.data` protected? Or any of the other ways to change the child nodes of a `script` element?

lukewarlow: The script element has an internal slot which won't match it's text content and so will go through the default policy.

annevk: So why would that not happen when `Node`'s setter is used? Seems like that would be fine.

lukewarlow: That would happen but if you don't have a default policy that's not web compatible. Because currently something that works will stop working.

annevk: Sure, but it's somewhat involved to actually do that, so the likelihood of that happening is small.

lukewarlow: > so the likelihood of that happening is small

Okay yeah fair, to be more accurate, it's a web compatibility risk.

I'm personally okay with this as a change if it makes the spec and implementation easier, and it's probably fine compat wise?

I'll update this PR to do it that way instead.

lukewarlow: Taking a look into how the IDL will look for this and then I'll push a commit up here that should hopefully make both the integration AND the spec much simpler at the cost of some potential edge cases if people's existing code is doing wonky stuff.

lukewarlow:
Actually based on the above discussion the existing IDL is correct and it's Chromium's implementation that is now not matching the spec. Going to go ahead and close this PR out. Will update any test expectations accordingly as part of https://github.com/w3c/trusted-types/issues/494 or any related implementation works.

lukewarlow:
@annevk hopefully this is now in a better shape. Shadows the properties onto HTMLScriptElement rather than messing with Node/textContent and Element/innerText. Uses the IDL attribute where possible so algorithms are mostly just setting the internal slot and calling the parent steps (Not sure how is correct way to do that apologies if it's wrong).

HTMLScriptElement/textContent uses a union type instead of the IDL extended attribute to avoid issues with nullable IDL type handling.

spec/index.bs
```diff
@@ -1186,36 +1186,62 @@ partial interface ShadowRoot {
...
+::  A string, containing the body of the script to execute that was set
+    through a {{StringContext}} compliant sink. Equivalent to script's
+    [=child text content=]. Initially null.
+
+#### The {{HTMLScriptElement/innerText}} IDL attribute #### {#the-innerText-idl-attribute}
+
+The {{HTMLScriptElement/innerText}} setter steps are:
+
+1.  Set {{HTMLScriptElement/[[ScriptText]]}} internal slot value to |value|.
+1.  Run the {{HTMLElement/innerText}} setter steps with the given value |value|.
```
lukewarlow, line 1218: Apologies if this isn't the correct way to call "parent" algorithms, they didn't seem to be linkified in their respective steps so couldn't do [=innerText setter steps=] for example.

annevk: That's correct, you have to abstract them first and then you'd be able to do that.

lukewarlow: Yeah that makes sense. My gut is that because this is aimed to be upstreamed very soon it's probably not worth worrying about the semantics too much as long as it's kinda makes sense?

lukewarlow: Ah that's true I missed that detail.

lukewarlow: https://github.com/whatwg/dom/pull/1271 exposes the Node/textContent getter and setter.

lukewarlow: https://github.com/whatwg/html/pull/10247 exposes Element/innerText getter and setter steps.

lukewarlow:
Updated this to link to the correct steps in the DOM and html specs so it should read better.

spec/index.bs
```diff
@@ -1137,36 +1143,78 @@ type policy factory]].
...
-This document modifies how {{HTMLScriptElement}} [=child text content=] can be set to allow applications to control dynamically created scripts. It does so by
-adding the {{HTMLElement/innerText}} and {{Node/textContent}} attributes directly on {{HTMLScriptElement}}. The behavior of the attributes remains the same
-as in their original counterparts, apart from additional behavior triggered by the {{StringContext}} extended attribute presence.
+The {{HTMLScriptElement/innerText}} setter steps are:
 
-Note: Using these IDL attributes is the recommended way of dynamically setting URL or a text of a script. Manipulating attribute nodes or text nodes directly will call a default policy on the final value when the script is prepared.
+1.  Let |value| be the result of calling [$Get Trusted Type compliant string$] with
+    {{TrustedScript}}, [=this=]'s [=relevant global object=], the given value, `HTMLScriptElement innerText`, and
+    `script`.
+1.  Set [=script text=] value to |value|.
```
annevk, line 1177: this's script text*

spec/index.bs
```diff
@@ -1137,36 +1143,78 @@ type policy factory]].
...
+
+#### The {{HTMLScriptElement/textContent}} IDL attribute #### {#the-textContent-idl-attribute}
+
+The {{HTMLScriptElement/textContent}} setter steps are to, if the given value is null, act as if it was the
+empty string instead, and then do as described below:
+
+1.  Let |value| be the result of calling [$Get Trusted Type compliant string$] with
+    {{TrustedScript}}, [=this=]'s [=relevant global object=], the given value, `HTMLScriptElement textContent`, and
+    `script`.
+1.  Set [=script text=] value to |value|.
```
annevk, line 1192: this's script text

spec/index.bs
```diff
@@ -1137,36 +1143,78 @@ type policy factory]].
...
+1.  Return the result of running [=get text content=] with [=this=].
+
+#### The {{HTMLScriptElement/text}} IDL attribute #### {#the-text-idl-attribute}
+
+Update the {{HTMLScriptElement/text}} setter steps algorithm as follows.
+
+1.  <ins>Let |value| be the result of calling [$Get Trusted Type compliant string$] with
+    {{TrustedScript}}, [=this=]'s [=relevant global object=], the given value, `HTMLScriptElement text`, and
+    `script`.</ins>
+1.  <ins>Set [=script text=] value to the given value.</ins>
```
annevk, line 1206: this's*

spec/index.bs
```diff
@@ -1137,36 +1143,78 @@ type policy factory]].
...
+1.  [=String replace all=] with the given value within this <code>script</code> element.
```
annevk, line 1207: just "this"? Seems weird to use this in step 1 and this script element here.

lukewarlow:
@koto and @annevk  does this seem ready to merge into the spec now?

spec/index.bs
```diff
@@ -1137,36 +1143,78 @@ type policy factory]].
...
+
+#### The {{HTMLScriptElement/text}} IDL attribute #### {#the-text-idl-attribute}
+
+Update the {{HTMLScriptElement/text}} setter steps algorithm as follows.
+
+1.  <ins>Let |value| be the result of calling [$Get Trusted Type compliant string$] with
+    {{TrustedScript}}, [=this=]'s [=relevant global object=], the given value, `HTMLScriptElement text`, and
+    `script`.</ins>
+1.  <ins>Set [=this=]'s [=script text=] value to the given value.</ins>
+1.  [=String replace all=] with the given value within this.
```
annevk, line 1207: xref this

annevk (commented):
I think I have the same question as I have for the HTML PR. Do we need to store a redundant copy of the text, or is the actual model invalidation of some kind of trusted bit.

lukewarlow:
@annevk 

Currently: We validate during prepare the text so we do need to store the text. The "invalidation" is currently only something we propose for the parsing behaviour.

It's possible that if all the ways to change the children of a script element end up firing children changed steps, and we get the neccessary changes there (still need to open a DOM PR) that we can do away with the script text entirely and just use invalidation and a trusted boolean "slot".

lukewarlow:
Raised #507 to discuss the "slot" vs invalidation model. Aside from that do the changes to the IDL here make sense?

annevk:
I think so.

lukewarlow:
I'm going to go ahead and get this merged, we'll get another chance when I upstream it to HTML to double check it all makes sense within the wider context of the full script enforcement.

mmarcell05 (commented):
Yet

Issue #492, opened 2024-03-27
Get trusted type compliant attribute value sink


lukewarlow:
https://w3c.github.io/trusted-types/dist/spec/#validate-attribute-mutation - This function specifies that the sink value when calling `document.createElement('iframe').setAttribute('srcdoc', 'foo');` should be `HTMLIframeElement srcdoc` the same as if you did `document.createElement('iframe').srcdoc = 'foo';`. However, this doesn't match Chromiums behaviour (where the sink is just `Element setAttribute`.

What should the spec say here? Should it just use the function as the sink (matching implementation)?

cc @koto  as he wrote this spec piece.

lukewarlow:
I've checked with Koto and the spec is intentionally different here. cc @otherdaniel so he's aware or the difference.

PR #493, opened 2024-03-28
Make sink an argument to get tt compliant attribute value

lukewarlow:
Partly address #492

***
<a href="https://pr-preview.s3.amazonaws.com/lukewarlow/trusted-types/pull/493.html" title="Last updated on Mar 28, 2024, 10:50 AM UTC (2b9be73)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trusted-types/493/39cc82e...lukewarlow:2b9be73.html" title="Last updated on Mar 28, 2024, 10:50 AM UTC (2b9be73)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1081,24 +1081,23 @@ Given an {{HTMLScriptElement}} (|script|), this algorithm performs the following
     If the algorithm threw an error, rethrow the error.
 
 ## Get Trusted Types-compliant attribute value ## {#validate-attribute-mutation}
-To <dfn abstract-op export>get Trusted Types-compliant attribute value</dfn> on {{Attr}} |attribute| with {{Element}} |element| and {{TrustedType}} or a string |newValue|, perform the following steps:
+To <dfn abstract-op export>get Trusted Types-compliant attribute value</dfn> on {{Attr}} |attribute| with {{Element}} |element|, {{TrustedType}} or a string |newValue|, and string |sink|, perform the following steps:
 
 1. Let |expectedType| be null.
-1. Let |sink| be null.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1081,24 +1081,23 @@ Given an {{HTMLScriptElement}} (|script|), this algorithm performs the following
     If the algorithm threw an error, rethrow the error.
 
 ## Get Trusted Types-compliant attribute value ## {#validate-attribute-mutation}
-To <dfn abstract-op export>get Trusted Types-compliant attribute value</dfn> on {{Attr}} |attribute| with {{Element}} |element| and {{TrustedType}} or a string |newValue|, perform the following steps:
+To <dfn abstract-op export>get Trusted Types-compliant attribute value</dfn> on {{Attr}} |attribute| with {{Element}} |element|, {{TrustedType}} or a string |newValue|, and string |sink|, perform the following steps:
```
lukewarlow, line 1084: Instead of the sink being HTMLIFrameElement srcdoc it will be something like `Element setAttribute` or `Attr value` this matches Chrome's behaviour.

lukewarlow: Need to double check which is correct though does Chrome's impl need updating or is the spec currently just wrong in this case?

lukewarlow: Though this change does mean the DOM integration needs to wire through the sinks from the IDL all the way down to the call out to this spec?

lukewarlow:
Matching commit to the DOM spec PR https://github.com/whatwg/dom/pull/1268/commits/656508956f278bf8178656a4f094850899822c7a

PR #497, opened 2024-04-04
Fix type issue between get tt compliant string and validate string in context

### LINKED ISSUE #488: "Validate the string in context" takes any value and calls "Get Trusted Type compliant string" which requires a TrustedType or a string
https://w3c.github.io/trusted-types/dist/spec/#html-validate-the-string-in-context

https://w3c.github.io/trusted-types/dist/spec/#get-trusted-type-compliant-string-algorithm

Should https://w3c.github.io/trusted-types/dist/spec/#html-validate-the-string-in-context try to stringify the value before passing it to https://w3c.github.io/trusted-types/dist/spec/#get-trusted-type-compliant-string-algorithm?

In any case, when https://w3c.github.io/trusted-types/dist/spec/#html-validate-the-string-in-context 's `value` has no stringifier, https://w3c.github.io/trusted-types/dist/spec/#html-validate-the-string-in-context should presumably throw an exception.

lukewarlow:
Get TT compliant string now takes any value, rather than just string or a TT. It then correctly stringifies this value.

Fixes #488

***
<a href="https://pr-preview.s3.amazonaws.com/lukewarlow/trusted-types/pull/497.html" title="Last updated on Apr 4, 2024, 1:03 PM UTC (c887020)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trusted-types/497/39cc82e...lukewarlow:c887020.html" title="Last updated on Apr 4, 2024, 1:03 PM UTC (c887020)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1022,19 +1022,21 @@ This algorithm will return a string that can be used with an
 It will ensure that the Trusted Type [=enforcement=] rules were respected.
 
 Given a {{TrustedType}} type (|expectedType|), a [=realm/global object=] (|global|),
-{{TrustedType}} or a string (|input|), a string (|sink|) and a string (|sinkGroup|), run these steps:
+an [=ECMAScript language value=] (|input|), a string (|sink|) and a string (|sinkGroup|), run these steps:
 
-1.  If |input| has type |expectedType|, return stringified
-    |input| and abort these steps.
+1.  Let |stringifiedInput| be the result of executing [$ToString$] on |input|.
+1.  If the algorithm threw an error, rethrow the error and abort the following steps.
+1.  If |input| has type |expectedType|, return |stringifiedInput| and abort these steps.
 1.  Let |requireTrustedTypes| be the result of executing [$Does sink type require trusted types?$] algorithm,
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1022,19 +1022,21 @@ This algorithm will return a string that can be used with an
 It will ensure that the Trusted Type [=enforcement=] rules were respected.
 
 Given a {{TrustedType}} type (|expectedType|), a [=realm/global object=] (|global|),
-{{TrustedType}} or a string (|input|), a string (|sink|) and a string (|sinkGroup|), run these steps:
+an [=ECMAScript language value=] (|input|), a string (|sink|) and a string (|sinkGroup|), run these steps:
```
lukewarlow, line 1025: If this now takes an ECMAScript language value which is what it's passed from the validate algorithm which gets it from WebIDL, does that "just work" for all the other uses of this algorithm? Or does it as I suspect break those call sites?

lukewarlow:
Going to go ahead and close this because we're going ahead with removing the extended attribute.

PR #499, opened 2024-04-04
Correctly set Script element's internal slot when parsed.

lukewarlow:


spec/index.bs
```diff
@@ -1178,6 +1179,30 @@ On setting the {{HTMLElement/innerText}}, {{Node/textContent}} and {{HTMLScriptE
...
+This document modifies the HTML parser to set the {{HTMLScriptElement/[[ScriptText]]}} internal slot value when the script is created.
+
+Modify the [=The text insertion mode=] algorithm as follows:
+
+<dl class="switch">
+    <dt id="scriptEndTag">An end tag whose tag name is "script"</dt>
+    <dd>
+        <p>...</p>
+
+        <ins>Set <var>script</var>'s <a href="#slots-with-trusted-values">[[ScriptText]]</a> internal slot value to its
```
lukewarlow, line 1193: This is too naive as the script content can be edited by script during parsing and so can't always be trusted.

We should hook into the https://dom.spec.whatwg.org/#concept-node-children-changed-ext children changed steps for the script element, to set a flag which if true means we can't trust it's contents and thus shouldn't set the internal slot.

lukewarlow, line 1193: We probably should also check that the internal slot isn't already set, if it is and we reach here (not even sure how it would happen?) we should probably leave it alone or maybe even reset to an empty string?

This way it ensures in the event of shenanigans that the spec plays it safe and assumes untrusted

lukewarlow: So both Chromium and WebKit not sure about Gecko, include an argument to their children changed steps to say what the source of the change was, so we could go `if changeSource is API then do a thing.` but the DOM spec doesn't define any such arguments.

@annevk I'd be interested to hear your thoughts on the best approach here.

The TLDR is that we can't rely on the text contents at the parser end step actually being from the parser. We need some way to determine that you haven't modified the script from JS while it's parsing (which is unfortunately possible). Any ideas on the best spec mechanisms for this?

Also cc @otherdaniel as you may have thoughts from your chromium implementation.

otherdaniel: > The TLDR is that we can't rely on the text contents at the parser end step actually being from the parser. We need some way to determine that you haven't modified the script from JS while it's parsing (which is unfortunately possible). Any ideas on the best spec mechanisms for this?
> 
> Also cc @otherdaniel as you may have thoughts from your chromium implementation.

We originally tried to make use of [parser-inserted](https://html.spec.whatwg.org/#parser-inserted), until we figured out that that doesn't quite mean what we thought it means. IMHO it'd be best if the [HTML script processing model](https://html.spec.whatwg.org/#script-processing-model) would maintain a "text content was parser inserted" flag which we could use, instead of us working around that from within the spec.

An alternative might be that the parser, at parse finish time, could pass along the text it actually inserted. (Which would then usually be the same as the one in the actual element.)


I take it the whole thing will become more a lot more complicated if streaming parsing (https://github.com/whatwg/html/issues/2142) actually materializes, which IMHO supports the case of having the parser keep track of this. (I haven't followed that discussion, so I'm not sure what its state is. The issue seems to see some activity.)

annevk: Are we concerned about `document.write()`?

@otherdaniel is it implemented through invalidation or the parser keeping track of what it inserted and comparing later? Because these are not equivalent. In particular if you modify but keep the text identical invalidation would have you not execute the script, but comparing later would.

lukewarlow:
```c++
void HTMLScriptElement::ChildrenChanged(const ChildrenChange& change) {
  HTMLElement::ChildrenChanged(change);
  if (change.IsChildInsertion())
    loader_->ChildrenChanged();

  // We'll record whether the script element children were ever changed by
  // the API (as opposed to the parser).
  children_changed_by_api_ |= !change.ByParser();
}
```

Here's the specific code that sets the flag. Afaict WebKit has a similar ChildChanged param to determine the source of the change, however children changed steps in the spec have no such param.

annevk: I guess we need to add it. I wonder why WebKit has it. I hope it's just for optimizations and not anything normative the specification currently doesn't cover.

lukewarlow: MIght be completely wrong but I wonder if some of this is related to the post insertion steps stuff that's been being discussed. Because the object also tells you what type of change happens which seems to be branched on a bit. (More than the source of the change)

annevk: I recommend filing an issue against DOM and describe the need you have and what you found and then we can copy the people working on that and see if there's some common ground.

(I suppose if you can tell ahead of time that something will invoke script you might have to do less calling around in the other cases, but not sure it's worth adding that to the model.)

lukewarlow: So coming back to this because I've been doing the implementation in WebKit I'm unable to get the test case to trigger untrusted script execution in any browser except Chrome (once I've disabled the TT enforcement). So I'm not even sure if this extra special case mechanism needs to be specced?

@otherdaniel in your tests were you able to get it working in other browsers?

I'm also not sure if this mechanism needs applying to SVG script elements too? It doesn't seem to be in CHromium?

annevk: If there's a difference between browsers there's a problem of sorts, no? Can you provide some more details?

lukewarlow: The test that triggers deferred parsing of scripts which allows you to have a reference before it's finished parsing, only works in Chrome. Within Firefox and Safari you never get access to the script element before it's finished parsing. I've tried to adjust the parameters of the test quite a bit and so far it's not worked. So I'm not sure if this issue is a chrome specific one or if this test is very heavily reliant on under specified behaviour which can lead to these differences.

annevk: I see. @domfarolino has been looking at this and might know. It would help to have a reference to the test or at least the code.

lukewarlow: https://github.com/web-platform-tests/wpt/blob/master/trusted-types/HTMLScriptElement-internal-slot.html

https://wpt.live/trusted-types/HTMLScriptElement-internal-slot.html

annevk: It seems weird that wouldn't reproduce in Gecko and WebKit. What if the new `<script>` element was inserted using `document.write()`? Presumably you could get a reference to it then, even if it hadn't executed yet. Because if you can't I think there's something else amiss. At least we're not observing defined behavior at that point.

lukewarlow: I've managed to get a test case that uses document.write. I thought that would be tricky because that's a sink itself, but I've made a default policy that no-ops for html.

I'll get a PR together that addresses this in WebKit and then I'll raise that issue against the DOM spec to see if we can get the necessary information in the children changed steps algorithm.

lukewarlow: I've raised a DOM issue https://github.com/whatwg/dom/issues/1288

lukewarlow: I've added an inline issue for this.

lukewarlow: For now I'm going to leave this out because that's partly caused by the API stuff which I'm defering to a follow up.

lukewarlow: Continue discussion in #525

PR #514, opened 2024-05-14
Replace EnsureCSPDoesNotBlockStringCompilation section with link to upstream PR

lukewarlow:
Also remove violation object changes as they're included in that PR


koto:
Do you want to hold on merging of this? For now the PR is in a draft status -- perhaps it makes sense to reference it, but not remove the algorithm here until the merge happens?

PR #518, opened 2024-05-28
Add an |includeReportOnlyPolicies| boolean argument to Does sink type require trusted types?

lukewarlow:
This is needed by https://github.com/w3c/webappsec-csp/pull/665

***
<a href="https://pr-preview.s3.amazonaws.com/lukewarlow/trusted-types/pull/518.html" title="Last updated on Jan 8, 2025, 6:23 PM UTC (80a006f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trusted-types/518/3819963...lukewarlow:80a006f.html" title="Last updated on Jan 8, 2025, 6:23 PM UTC (80a006f)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1033,7 +1033,7 @@ Given a {{TrustedType}} type (|expectedType|), a [=realm/global object=] (|globa
 1.  If |input| is an instance of |expectedType|, return stringified
     |input| and abort these steps.
 1.  Let |requireTrustedTypes| be the result of executing [$Does sink type require trusted types?$] algorithm,
-    passing |global|, and |sinkGroup|.
+    passing |global|, |sinkGroup|, and true.
 1.  If |requireTrustedTypes| is `false`, return stringified |input| and abort these steps.
 1.  Let |convertedInput| be the result of executing [$Process value with a default policy$] with the same arguments as this algorithm.
 1.  If the algorithm threw an error, rethrow the error and abort the following steps.
@@ -1399,11 +1399,9 @@ Content-Security-Policy: trusted-types one two default
 
 ### <dfn abstract-op>Does sink type require trusted types?</dfn> ### {#does-sink-require-trusted-types}
 
-Given a [=realm/global object=] (|global|), a string (|sinkGroup|) this algorithm
-returns `true` if the [=injection sink=] requires a [=Trusted Type=], and
-`false` otherwise.
+Given a [=realm/global object=] (|global|), a string (|sinkGroup|) and a boolean (|includeReportOnlyPolicies|), this algorithm
+returns `true` if the [=injection sink=] requires a [=Trusted Type=], and `false` otherwise.
 
-1.  Let |result| be `false`.
 1.  For each |policy| in |global|'s <a>CSP list</a>:
     1.  If |policy|'s <a>directive set</a> does not contain a <a>directive</a>
         whose [=directive/name=] is `"require-trusted-types-for"`, skip to the next |policy|.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1411,19 +1411,19 @@ Content-Security-Policy: trusted-types one two default
...
 1.  For each |policy| in |global|'s <a>CSP list</a>:
     1.  If |policy|'s <a>directive set</a> does not contain a <a>directive</a>
         whose [=directive/name=] is `"require-trusted-types-for"`, skip to the next |policy|.
     1.  Let |directive| be the |policy|'s |directive set|'s [=directive=] whose name
         is `"require-trusted-types-for"`
     1.  If |directive|'s [=directive/value=] does not contain a <a>trusted-types-sink-group</a> which is a match
         for |sinkGroup|, skip to the next |policy|.
-    1.  Set |result| to `true`.
+    1.  Let |enforced| be true if |policy|'s [=policy/disposition=] is `"enforce"`, and false otherwise.
+    1.  Set |result| to true if |enforced| is true, and set to |includeReportOnly| otherwise.
```
fred-wang, line 1426: If I read this correctly, the name `includeReportOnly` seems wrong, since the "enforce" policies are always forcing the result to true ; while the "report" policies only forces the result to true if `includeReportOnly` is false.

Also the "set to |includeReportOnly|" wording sounds a bit ambiguous to me. At least as a non-native speaker I initally misunderstood this meant setting the result variable to the value of includeReportOnly, which does not make sense.

So I guess what you want is rename the parameter `includeEnforceOnly` and rewrite the line to something like:

1. If `includeEnforceOnly` is false or |policy|'s [=policy/disposition=] is `"enforce"` then set |result| to `true`.

The boolean values of the caller would still need to be changed to includeEnforceOnly=false if we want to keep including all the policies.

Finally, even before your change it seems the result can only be set to true and never goes back to false ; so probably you can either just exit the "for each" loop immediately when it's set to true. Alternatively, remove the "result" variable and use "return true" or "return false" statements.

spec/index.bs
```diff
@@ -1018,7 +1018,7 @@ Given a {{TrustedType}} type (|expectedType|), a [=realm/global object=] (|globa
 1.  If |input| has type |expectedType|, return stringified
     |input| and abort these steps.
 1.  Let |requireTrustedTypes| be the result of executing [$Does sink type require trusted types?$] algorithm,
-    passing |global|, and |sinkGroup|.
+    passing |global|, |sinkGroup|, and true.
```
fred-wang, line 1021: So with includeReportOnly=true means we change current behavior and only include policies with "report" disposition. That does not seem to be what you want?

fred-wang (changes requested):

spec/index.bs
lukewarlow: So this method right now returns true if there's a `require-trusted-types-for` CSP header. Regardless of whether it's report only or not. Because currently it's used to go down the default policy codepath etc.

However, with the new CSP script-src keyword, we don't want that. We want this to only return true if it is *enforced*.

So this is updated with a new |includeReportOnly| boolean, which is set to true in the current usage, and set to false in the CSP PR.

So this then returns true if the policy is enforced, and additionally returns true if the policy is report only and |includeReportOnly| is true.

I'll reword it a bit to hopefully make it clearer, and add the early return when the result is true.

lukewarlow: I've realised the ambiguity regarding the variable name too. Good point, I've changed it to includeReportOnlyPolicies to hopefully make it clearer what is meant.

fred-wang (approved):
Thanks, that much clearer with the "includeReportOnlyPolicies" name.

Issue #517, opened 2024-05-20
Should all 3 script IDL setters change the associated script text value identically


lukewarlow:
A good point came up during code review of an associated webkit patch that the .innerText setter steps and the .textContent/.text setter steps are different, and presumably could result in a different child contents value? If this is the case the current spec (and Chromium implementation) could lead to some script execution errors despite being set via a trusted object and sanctioned IDL attribute?

Question 1: Are these actually identical in behaviour?

Question 2: If they're not should the spec and implementationed be updated to account for the differences?

annevk:
Oh yeah, `innerText` is wild: https://software.hixie.ch/utilities/js/live-dom-viewer/?%3Cscript%3E%0Adocument.querySelector(%60script%60).innerText%20%3D%20%60test%5Cntest%60%0A%3C%2Fscript%3E

Can we just not support that?

lukewarlow:
cc @otherdaniel and @koto curious what your thoughts on this are. I suspect that people might be using this in the wild? Question is can we remove it as a valid IDL sink *or* do we just document in the spec that it's an oddity and might not always work as expected?

koto:
> [setting script content via innerText] could lead to some script execution errors despite being set via a trusted object

Isn't this the case even without Trusted Types? The spec doesn't provide any indication that the value will be functional for any given sink, only that the value has passed through a policy so there's no DOM XSS risk with assigning it.

> Can we just not support that?

We definitely need to cover the DOM XSS sinks with Trusted Types, and script contents being set via innerText is used commonly in applications. TT consciously tries to work with the existing sinks, as opposed to e.g. proposing new ones and disabling old ones.

lukewarlow:
> > [setting script content via innerText] could lead to some script execution errors despite being set via a trusted object

> Isn't this the case even without Trusted Types? The spec doesn't provide any indication that the value will be functional for any given sink, only that the value has passed through a policy so there's no DOM XSS risk with assigning it.

Specifically I mean trusted types related script execution errors. Because the value may mismatch in which case TT will reject it, which could be unexpected.

Sounds like we should either update the implementations and spec to set the value *after* it's been set so that it matches. Or just to add a note that explains there may be a mismatch?

annevk:
Commonly? Can you point to an example? Given what it does with newlines I'd expect all kinds of brokenness.

koto:
[...]

PR #524, opened 2024-06-12
Add script text associated data to SVGScriptElement.

lukewarlow:
Addresses #483


spec/index.bs
```diff
@@ -1175,14 +1175,24 @@ This document modifies {{HTMLScriptElement}}s. Each script has:
...
     [=child text content=]. Initially an empty string.
 
+This document also modifies {{SVGScriptElement}}s. Each script has:
+
+: an associated string <dfn export for="SVGScriptElement">script text</dfn>.
+::  A string, containing the body of the script to execute that was set
+    through a compliant sink. Equivalent to script's
+    [=child text content=]. Initially an empty string.
+
+Issue: There's currently no defined processing model for {{SVGScriptElement}}s. This spec assumes one that is similar to that of {{HTMLScriptElement}}'s model.
```
lukewarlow, line 1185: This feels iffy but there really seems to be no spec for SVGScriptElement's atm. So this is an attempt to say "use common sense and good luck"

lukewarlow: There's https://www.w3.org/TR/SVGMobile12/script.html#ScriptContentProcessing which is an old version of an SVG spec which could be monkey patched but idk that it really provides much value doing that.

lukewarlow:
I've rebased this atop the parser PR so I can account for everything SVG in one PR.

spec/index.bs
```diff
@@ -1224,6 +1236,28 @@ The {{HTMLScriptElement/src}} setter steps are:
     `script`.</ins>
 1.  <ins>Set [=this=]'s [=src=] content attribute to |value|.</ins>
 
+#### Setting slot values from parser #### {#setting-slot-values-from-parser}
+
+This document modifies the HTML parser to set the [=HTMLScriptElement/script text=] value when the script is created.
+
+Modify the [=The text insertion mode=] algorithm as follows:
```
lukewarlow, line 1243: Need to work out if this only for HTMLScriptElement's or if it somehow applies to SVG too.

annevk:
Wouldn't it be better to handle HTML and SVG script at the same time?

lukewarlow:
> Wouldn't it be better to handle HTML and SVG script at the same time?

Perhaps but they're pretty distinct from each other spec wise. This PR adds the "script text" and parser handling to svg script element that HTML script has. The only bit this PR is missing is actually doing the enforcement. I'm not sure how best to go about that due to the fact that SVG doesn't have a "prepare the script" method definition.

#525 will account for the more complicated parser stuff in one go for both SVG and HTML.

spec/index.bs
```diff
@@ -1270,10 +1298,12 @@ The first few steps of the [=prepare the script element=] algorithm are modified
...
     updates it, it will execute in an async fashion even if the <code id=script-processing-model:attr-script-async-5><a href=https://html.spec.whatwg.org/#attr-script-async>async</a></code> attribute isn't set.</p>
   <li><ins><p>Execute the [$Prepare the script text$] algorithm on <var>el</var>. If that algorithm threw an error, then return.</p></ins></li>
-  <li><p>Let <var ignore="">source text</var> be <var>el</var>'s <del><a id=script-processing-model:child-text-content href=https://dom.spec.whatwg.org/#concept-child-text-content data-x-internal=child-text-content>child text content</a>.</del> <ins>[=script text=] value.</ins>
+  <li><p>Let <var ignore="">source text</var> be <var>el</var>'s <del><a id=script-processing-model:child-text-content href=https://dom.spec.whatwg.org/#concept-child-text-content data-x-internal=child-text-content>child text content</a>.</del> <ins>[=HTMLScriptElement/script text=] value.</ins>
   <li>...
   </ol>
 
+Issue: There's no proper definition for the processing of SVG script elements. However, you should apply a similar change to the processing of {{SVGScriptElement}}s.
```
lukewarlow, line 1305: @annevk can you think of a better way to word this, or a way to actually spec this?

I could try patching https://www.w3.org/TR/SVGMobile12/script.html#ScriptContentProcessing but it's a bit lose on the details and is fairly old, so idk that we'd ever upstream the patching.

lukewarlow: The boolean bit is currently only used as part of the "have the children been changed by an API during parsing" code.

lukewarlow: The problem with a Boolean bit is how to allow the specific IDL and the "actually parsed" stuff but nothing else. I'm currently unable to think how you would do that.

annevk: Everything else that mutates would end up flipping the bit? It might help if we went through all the entry points we are concerned about or why this would be hard for some of them.

annevk: E.g., is the tradeoff that some mutation algorithm would need to have knowledge about script elements? And what do implementations do here? I find it hard to believe they would accept doubling the cost of each script element.

lukewarlow: So we want to allow when the parser gets to `</script>` and the contents haven't already been touched by JS.

We also want to allow HTMLScriptElement.text HTMLScriptElement.textContent and HTMLScriptElement.innerText to allow setting the contents (once the script is parsed).

So examples of what we want to fail are stuff like appendChild(textNode), or Node.textContent setter being accessed weirdly with the script element.

So is your idea to update basically every callsite with an "if script flip trusted bit to false"? I think my worry is that we'd want that change to be deeper than callsites and then it's a matter of how to say this mutation is okay and this one isn't.

Chromium right now does the store a second copy of contents and only touches the 3 IDL setters, the parser end step and children changed steps.

lukewarlow: I'm a bit confused by the question so I think I'm misunderstanding something here.

Yes we want to track if it's an API or parser triggering children changed steps. So currently this is used for the weird API change mid-parse case.

If we used a boolean for trust rather than duplicating the internals we could mark is as trusted in parser and that would be fine.

But after parse is finished this would be triggered by all API changes most of which we don't want to trust. But it would also be triggered by HTMLScriptElement.text for example. How would we allow these sanctioned APIs?

If we do HTMLScriptElement.text setter sets the trusted bool then that's fine but it would then trigger children changed steps as an API and we'd have to handle that (presumably the idea would be to mark it as untrusted?)

annevk: In the cases where we trust the API can't we just set the bit to trusted after the children changed steps have run? I suppose with mutation events that might not quite work. But they're close to disappearing. Hmm.

lukewarlow: Yeah I did consider something like that where we have a boolean to say this is an approved API change and then in children changed we use that to set trusted (so we'd have maybe 3 booleans?), but I felt slightly uneasy about it.

It's possible though that in practice any shenanigans would result in an untrusted script.

annevk: I'm still at 1 boolean. Presumably if a script set is trustworthy you can just set the existing bit to trustworthy in the end.

lukewarlow: Ah yeah actually we don't need a parse specific one anymore and provided it's all synchronous you're right you could set the "trusted" bit to true after the children changed steps (which would set it to false for all API calls).

If we make children changed always untrust API changes I think that solves any weirdness from mutation events too. Because we'd fail safe and the script would be untrusted. Because we only the 3 sanctioned APIs to behave properly wrt to mutation events.

Thinking on it some more I think that could work?

cc @otherdaniel what do you think about that (Assuming we can get children changed steps spec to correctly tell us if it's API or parser)?

Relatively small changes needed (assuming my mental model is right) such as that below and then updating the children changed steps and the prepare script steps:

image

lukewarlow: Thinking on this some more it would also fix https://github.com/w3c/trusted-types/issues/517 (different handling of new lines leading to TT errors)

lukewarlow: Okay so I tried implementing this and hit a snag.

https://html.spec.whatwg.org/#prepare-the-script-element:~:text=When%20a%20script%20element%20el%20that%20is%20not%20parser%2Dinserted%20experiences

This bit of the HTML spec is ran inside of children changed steps (at least in WebKit and I believe Chromium from a quick look), so prepare the script runs *before* the isTrusted flag is set back to true at the end of setInnerText.

lukewarlow: I've opened a new draft PR which I'll use to do this new model (and will also apply to svg as best it can) https://github.com/w3c/trusted-types/pull/533

Issue #472, opened 2024-03-07
Adopt Infra syntax throughout


annevk:
E.g., in Create a Trusted Type Policy it would be much clearer if the dictionary members were set using map syntax. It currently states properties are being set, which is just not true.

PR #527, opened 2024-06-18
Handle null policyValue in Create a Trusted Type

### LINKED ISSUE #469: Stringification of TrustedHTML with `null`-data needs to be specified
https://w3c.github.io/trusted-types/dist/spec/#abstract-opdef-create-a-trusted-type

lukewarlow:
Fixes #469

***
<a href="https://pr-preview.s3.amazonaws.com/lukewarlow/trusted-types/pull/527.html" title="Last updated on Jun 18, 2024, 1:41 PM UTC (3e9fe32)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trusted-types/527/e776fc6...lukewarlow:3e9fe32.html" title="Last updated on Jun 18, 2024, 1:41 PM UTC (3e9fe32)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -981,6 +981,7 @@ a string |value| and a list |arguments|, execute the following steps:
 1.  Let |policyValue| be the result of executing [$Get Trusted Type policy value$] with the same arguments as this algorithm and additionally true as |throwIfMissing|.
 1.  If the algorithm threw an error, rethrow the error and abort the following steps.
 1.  Let |dataString| be the result of stringifying |policyValue|.
+1.  If |policyValue| is null or undefined, set |dataString| to the empty string.
 1.  Return a new instance of an interface with a type
     name |trustedTypeName|, with its associated data value set to |dataString|.
 


spec/index.bs
```diff
@@ -981,6 +981,7 @@ a string |value| and a list |arguments|, execute the following steps:
 1.  Let |policyValue| be the result of executing [$Get Trusted Type policy value$] with the same arguments as this algorithm and additionally true as |throwIfMissing|.
 1.  If the algorithm threw an error, rethrow the error and abort the following steps.
 1.  Let |dataString| be the result of stringifying |policyValue|.
```
lukewarlow, line 983: Not sure if stringifying is even defined to mean a specific thing, this spec in general probably needs checking over for this sort of thing but leave that for #472

mbrodesser-Igalia (approved):

spec/index.bs
```diff
@@ -981,6 +981,7 @@ a string |value| and a list |arguments|, execute the following steps:
...
+1.  If |policyValue| is null or undefined, set |dataString| to the empty string.
```
petervanderbeken, line 984: I don't understand how this solves the issue? Stringifying `policyValue` would still happen with `policyValue` potentially being `null`?

Seems to me like this should be
```
1. If |policyValue| is null or undefined, set |dataString| to the empty string. Otherwise let |dataString| be the result of stringifying |policyValue|.
```

lukewarlow: It stringifies it which would set dataString to "null" (presumably) and then dataString is set to an emptyString in the case the original value was null. Just feels cleaner to read imo.

petervanderbeken: Looking more into it, I find it quite confusing. I'm going to assume that **Get Trusted Type policy value** actually calls https://webidl.spec.whatwg.org/#invoke-a-callback-function, since it's technically invoking a WebIDL callback, not an ECMAScript function. After invoking the callback we have a nullable `DOMString` WebIDL value. Stringifying a value of that type is not specified anywhere afaik? Having `undefined` mentioned also doesn't make sense for a value of that type.

Isn't it enough to then just replace these 2 steps with:
```
1. If |policyValue| is null, set |policyValue| to the empty string.
```
and replace `dataString` with `policyValue` in the following step?

lukewarlow: Yeah we need to look more into the spec as a whole it's probably missing correct infra syntax in many places. This change was largely just to make it clear what to do in that specific case.

Cc @koto would you have any spare capacity to spend some time making the spec more correct from an editorial perspective?

Issue #525, opened 2024-06-12
Script element mid-parse protection mechanism


lukewarlow:
This issue is here to track the specific spec mechanism for protecting against a script that's edited mid-parse.

To avoid needing to look all over here's a summary. As of https://github.com/w3c/trusted-types/pull/499 all parsed scripts are trusted. The problem with this approach is that a script elements contents can be changed from JS before it finishes parsing. This is case tested by the first subtest of https://wpt.live/trusted-types/HTMLScriptElement-internal-slot.html and https://wpt.live/trusted-types/SVGScriptElement-internal-slot.html.

Chromium and WebKit's approach is to store a boolean flag on the HTML/SVGScriptElement which gets set to true, if the children of the script element are changed by an API. This flag is then checked before setting the script text data.

We cannot spec this directly due to the DOM specs children changed steps not providing the information as to whether something is changed by an API or parser.  https://github.com/whatwg/dom/issues/1288 - has been raised to discuss changes to the DOM spec which would be needed for this.

lukewarlow:
cc @mbrodesser-Igalia  it would be good if you or someone else from mozilla could provide input on whether Gecko has a similar concept to WebKit and Chromium here. If they do then it makes the case for standardising this, and specifically this protection mechanism better.

smaug----:
Gecko doesn't have such concept.

mbrodesser-Igalia:
> > Gecko doesn't have such concept.
> 
> To be explicit: it seems (https://searchfox.org/mozilla-central/search?q=children.*changed&path=&case=false&regexp=true) Gecko doesn't implement the children changed steps.

For the record: Gecko implements the children changed steps via `nsIMutationObserver` (https://searchfox.org/mozilla-central/rev/ece43b04e7baa4680dac46a06d5ad42b27b124f4/dom/base/nsIMutationObserver.h#110).

domfarolino:
> The problem with this approach is that a script elements contents can be changed from JS before it finishes parsing.

So one scenario in which a parser-inserted script should be considered untrusted is when it edits itself mid-parse, which I think is what the document.write test case highlights that you pointed to. I think another scenario would be if a parser-inserted script that cannot immediately execute (maybe it has an invalid `type` attribute) has its children mutated via an API before its `type` becomes valid again. This is also a case that our special boolean flag (should we ever add one) would have to capture, right?

[...]

Issue #507, opened 2024-04-22
Script element protection model


lukewarlow:
This issue is to discuss the model for protecting script elements, in the spec we have a mostly finalised idea for protecting them (except during parsing which needs work).

However, it currently relies on us storing a text representation of the children of the script element. If we could instead invalidate a "trusted" boolean when APIs are called this would be better.

For the parsing support there's discussion about modifiying children changed steps to pass through the source of the change, we could use this in general to invalidate the trusted bit (which would start as true).

This might not work as we'd like however as the IDL that should allow you to update would also trigger the children changed steps as if it were an API call?

lukewarlow:
Aside from the parsing concern I think the current model is simplest, though less memory efficient.

annevk:
As I noted in the HTML issue there's an observable difference as well. With one of the models if you set it to the same value it will still execute, but with the other it won't.

For trusted API callers you could manipulate the contents and then set the trusted bit (possibly after verifying the contents still match the input), maybe?

Sawsqr68:
Sorry, there are no children in the Council of Members or in the work team.. There are only fraudsters and impersonators!!!

PR #533, opened 2024-07-04
Change Script Enforcement Mechanism to use flags

lukewarlow:
Also add SVGScriptElement to spec


spec/index.bs
```diff
@@ -1179,10 +1177,19 @@ partial interface HTMLScriptElement {
 
 This document modifies {{HTMLScriptElement}}s. Each script has:
 
-: an associated string <dfn export for="HTMLScriptElement">script text</dfn>.
-::  A string, containing the body of the script to execute that was set
-    through a compliant sink. Equivalent to script's
-    [=child text content=]. Initially an empty string.
+: an associated boolean <dfn export for="HTMLScriptElement">is trusted</dfn>.
+::  A boolean indicating whether a script element is considered trustworthy for execution.
+    Initially true.
```
lukewarlow, line 1182: Initially true which covers scripts inline in the page being parsed.

spec/index.bs
```diff
@@ -1233,29 +1240,21 @@ The {{HTMLScriptElement/src}} setter steps are:
...
 1.  <ins>Set [=this=]'s [=src=] content attribute to |value|.</ins>
 
-#### Setting slot values from parser #### {#setting-slot-values-from-parser}
+#### Script children changed steps #### {#script-children-changed-steps}
 
-This document modifies the HTML parser to set the [=script text=] value when the script is created.
+This document modifies the [=children changed steps=] for {{HTMLScriptElement}} as follows:
 
-Modify the [=The text insertion mode=] algorithm as follows:
+1.  If <var ignore=''>parserChange</var> is false, set [=this=]'s [=HTMLScriptElement/is trusted=] to false.
```
lukewarlow, line 1247: This `parserChange` is a placeholder for what we end up speccing in https://github.com/whatwg/dom/issues/1288

spec/index.bs
```diff
@@ -1233,29 +1240,21 @@ The {{HTMLScriptElement/src}} setter steps are:
...
 
-<dl class="switch">
-    <dt id="scriptEndTag">An end tag whose tag name is "script"</dt>
-    <dd>
-        <p>...</p>
+1.  If [=this=]'s [=HTMLScriptElement/changed by trusted sink=] is true, set [=this=]'s [=HTMLScriptElement/is trusted=] to true.
```
lukewarlow, line 1249: This flag is used to say hey this is an API change but it's a trusted one. We unset the flag once used.

spec/index.bs
```diff
@@ -1233,29 +1240,21 @@ The {{HTMLScriptElement/src}} setter steps are:
...
 
-        <ins><p>Set <var>script</var>'s [=script text=] value to its [=child text content=].</p></ins>
+1.  Set [=this=]'s [=HTMLScriptElement/changed by trusted sink=] to false.
 
-        <p>If the <span>active speculative HTML parser</span> is null, then <span>prepare the script
-        element</span> <var>script</var>. This might cause some script to execute, which might cause
-        <span data-x="dom-document-write">new characters to be inserted into the tokenizer</span>, and
-        might cause the tokenizer to output more tokens, resulting in a [=reentrant invocation of the parser=].</p>
+Issue: Need to double check how [part of script element's spec](https://html.spec.whatwg.org/#prepare-the-script-element:~:text=When%20a%20script%20element%20el%20that%20is%20not%20parser%2Dinserted%20experiences) fits into this. These steps need to happen before prepare the script is called.
```
lukewarlow, line 1253: I think we need to change the html spec when upstreaming to run the prepare the script (under relevant conditions) at the end of the children changed steps.

lukewarlow: Interestingly it would rely on having a bit more information in the children changed steps algorithm if we want to inline it. Because it needs to know what type of change it is (insertion specifically in this case).

I suspect this is why some Chrome and WebKit's childrenChanged functions include more than the dom spec's algorithm. (And is why Firefox implements it in a way that also gives them this more granular informaion).

spec/index.bs
```diff
@@ -1179,10 +1163,19 @@ partial interface HTMLScriptElement {
 
 This document modifies {{HTMLScriptElement}}s. Each script has:
 
-: an associated string <dfn export for="HTMLScriptElement">script text</dfn>.
-::  A string, containing the body of the script to execute that was set
-    through a compliant sink. Equivalent to script's
-    [=child text content=]. Initially an empty string.
+: an associated boolean <dfn export for="HTMLScriptElement">is trusted</dfn>.
```
annevk, line 1166: I think `for` is either comma or vertical bar separated so you can define it for HTML and SVG `script` once.

spec/index.bs
annevk: @domfarolino should probably look at this.

Also would that create issues with re-entrant invocations?

spec/index.bs
```diff
@@ -1206,8 +1199,8 @@ empty string instead, and then do as described below:
 1.  Let |value| be the result of calling [$Get Trusted Type compliant string$] with
     {{TrustedScript}}, [=this=]'s [=relevant global object=], the given value, `HTMLScriptElement textContent`, and
     `script`.
-1.  Set [=this=]'s [=script text=] value to |value|.
 1.  Run [=set text content=] with [=this=] and |value|.
+1.  Set [=this=]'s [=HTMLScriptElement/changed by trusted sink=] to true.
```
annevk, line 1203: If as you suggest below "prepare the script" runs at the end of "set text content" it seems this wouldn't really work?

lukewarlow: Ah yeah sorry I forgot to change the order back for this one

spec/index.bs
mbrodesser-Igalia: Which issues, besides the one mentioned in https://github.com/w3c/trusted-types/pull/533#issue-2391021496, is this PR intended to fix?

lukewarlow: https://github.com/w3c/trusted-types/issues/525, and https://github.com/w3c/trusted-types/issues/507 I think should both be closable once this PR is finished alongside the SVG specific one.

spec/index.bs
mbrodesser-Igalia: Might be worth promoting that comment to a note in the spec.

spec/index.bs
mbrodesser-Igalia: When "parserChange" is `false` and `changed by trusted sink` is `true`, couldn't still malicious code have been injected? E.g. if a trusted sink called only `someScript.innerText = someScript.innerText` that'd make the untrusted code trusted.

lukewarlow: > E.g. if a trusted sink called only `someScript.innerText = someScript.innerText` that'd make the untrusted code trusted.

That would only work if a default policy had sanctioned that value. Else the assignment would fail before the "trusted sink" Boolean is set

spec/index.bs
lukewarlow: It would seem that https://github.com/whatwg/html/pull/10188 already changes that part of the HTML spec to be defined in terms of the children changed steps so I think we'd just need to put our new steps first and then run the post-insertion steps and it'll fix the concerns I had here.

lukewarlow:
@annevk one question I've got with this boolean approach. Does something like .innerText =`a/nb` cause multiple children changed steps invocations? Wondering if the `<br>` element generation might somehow mess things up?

lukewarlow:
If children changed steps are never called by the parser then I think I can just remove the new boolean and this PR is basically ready to go. I was under the impression they would be called by the parser but I don't think they are.

@domfarolino do you know off the top of your head if that's true? I know you've used them for the post connection steps work youve been doing

lukewarlow:
I've updated this PR to remove the parserChange boolean because it appears it's not needed. Spec wise the children changed steps seem to never actually get called by the parser. So parserChange is always false (aka it's always an API).

fred-wang (commented):
@lukewarlow I guess you are still waiting for a review? I checked the PR and it looks good to me, but I really think someone more familiar with the spec/code should review.

koto (approved):

spec/index.bs
```diff
@@ -1233,29 +1222,25 @@ The {{HTMLScriptElement/src}} setter steps are:
...
+1.  If [=this=]'s [=HTMLScriptElement/changed by trusted sink=] is true, set [=this=]'s [=HTMLScriptElement/is trusted=] to true.
 
-        <ins><p>Set <var>script</var>'s [=script text=] value to its [=child text content=].</p></ins>
+1.  Set [=this=]'s [=HTMLScriptElement/changed by trusted sink=] to false.
 
-        <p>If the <span>active speculative HTML parser</span> is null, then <span>prepare the script
-        element</span> <var>script</var>. This might cause some script to execute, which might cause
-        <span data-x="dom-document-write">new characters to be inserted into the tokenizer</span>, and
-        might cause the tokenizer to output more tokens, resulting in a [=reentrant invocation of the parser=].</p>
+Note: This relies on the children changed steps never being called by the parser.
```
domfarolino, line 1235: How can we assume this? I think these steps are called indirectly by the parser, in the usual insertion path, no? And it seems that browsers expect this too: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/html/html_script_element.cc;l=82;drc=d81683e439e0b80cac98e017d94449d59f48fa0d.

domfarolino (commented):
I know this has already landed but I took a very belated look and have at least one question

spec/index.bs
lukewarlow: Apologies for the slow response I've been OoO I didn't neccesarily expect this PR to be merged until someone with more knowledge of the status quo had reviewed, apologies for not making that clearer.

So from all I could work out, the specification children changed steps are only called from APIs and NOT the parser. 
If that's not true then this will need reverting. 
The ChildrenChanged method implemented in Chromium and WebKit are called by the parser but that's because they aren't a 100% match for the spec.

Fwiw I did a quick prototype in Chromium and it seemed to work as expected (it was a few months back at this point so can't fully remember). I've also got a draft PR to update WebKit to use this new model too https://github.com/WebKit/WebKit/pull/30649.

domfarolino: > Apologies for the slow response I've been OoO

Hah, not a problem, my delay since Anne pinged be about this was far worse.

> I didn't neccesarily expect this PR to be merged until someone with more knowledge of the status quo had reviewed, apologies for not making that clearer.

Do you think we should revert it?

> So from all I could work out, the specification children changed steps are only called from APIs and NOT the parser.
If that's not true then this will need reverting.

Hmm, spec-wise, I think the children changed steps are indeed called from the parser. At the very least, the HTML's adoption agency algorithm [creates an element](https://html.spec.whatwg.org/C#parsing-main-inbody:create-an-element-for-the-token) and then immediately appends it to the DOM, which goes through the https://dom.spec.whatwg.org/#concept-node-insert path, which itself [calls the children changed steps](https://dom.spec.whatwg.org/#ref-for-concept-node-children-changed-ext).

I am really not familiar with the HTML parser, so I can't find any other paths from parser -> DOM insertion hook (which calls children changed steps), but I'm pretty sure they exist. For example, when a script is run immediately after insertion to the DOM, it does so via the post-connection steps: https://html.spec.whatwg.org/C#script-processing-model:html-element-post-connection-steps. And the only thing that calls these steps [is the DOM insertion algorithm](https://dom.spec.whatwg.org/#ref-for-concept-node-post-connection-ext%E2%91%A5)—the same thing that calls the children changed steps—so it must be invoked during the parser. Same goes for iframes, which initialize in the same way.

lukewarlow:
Based on the above @koto could we revert this?

Cc @fred-wang

lukewarlow:
If we do revert this then we still have spec work required to actually finish this which I personally don't have capacity to continue but I'm happy to help others who wish to do that.

koto:
Yeah, we can revert this.

lukewarlow:
So yeah this comes back to us needing new data passed down the children changed steps to say whether it's a parser or API change. Idk how much work that is to do, and I don't know if it's compatible with a Mozilla implementation (as WebKit and chrome are quite similar in this respect)

Issue #553, opened 2024-10-22
getPropertyType/getAttributeType when null namespaces are passed.


fred-wang:
cc @smaug---- @mbrodesser-Igalia @lukewarlow 

`elementNs` and `attrNS` are optional nullable arguments defaulting to the empty string, so we should test what happens when null values are used:

```
optional DOMString? elementNs = ""
optional DOMString? attrNs = ""
```

- For `attrNS`, a null value is explicitly handled in  [Get Trusted Type data for attribute](https://w3c.github.io/trusted-types/dist/spec/#abstract-opdef-get-trusted-type-data-for-attribute). This case is covered by "getAttributeType with full namespace info" cases of `TrustedTypePolicyFactory-getAttributeType-namespace.html` because `attribute.namespaceURI` is null here, but maybe a more explicit test with a null argument would help.

- For `elementNs`, the step retrieving the "element interface for localName and elementNs" is similar to step 7 of [DOM's create an element](https://dom.spec.whatwg.org/#concept-create-element). Probably this is not very explicit, but I understand this falls back to the `Element` interface for null namespace, similarly to [createElementNs(null, "foo")](https://dom.spec.whatwg.org/#internal-createelementns-steps). However, we don't have any tests for that (see #429).

fred-wang:
> For elementNs, the step retrieving the "element interface for localName and elementNs" is similar to step 7 of [DOM's create an element](https://dom.spec.whatwg.org/#concept-create-element). Probably this is not very explicit, but I understand this falls back to the Element interface for null namespace, similarly to [createElementNs(null, "foo")](https://dom.spec.whatwg.org/#internal-createelementns-steps). However, we don't have any tests for that (see https://github.com/w3c/trusted-types/issues/429).

So checking on Chromium and WebKit, an explicitly null elementNs seems to fallback to the HTML namespace.

koto:
This seems to be resolved in the spec via https://github.com/w3c/trusted-types/pull/557, and in WPT via https://github.com/web-platform-tests/wpt/pull/48769. @fred-wang, do you agree this can be closed now?

fred-wang:
Right, we can close it now.

PR #557, opened 2024-10-23
Treat a null elementNs in getPropertyType/geAtributeType as the HTML …

fred-wang:
…namespace

This is the behavior that Chromium and WebKit currently implement, see https://phabricator.services.mozilla.com/D226547

Currently, it seems that the "interface for localName and elementNs" would fallback to `Element`, similarly to what happens with `document.createElementNS(null, "div")` for example.

Note that a null attrNs is already handled by the spec.

See also https://github.com/w3c/trusted-types/issues/553 for details.

***
<a href="https://pr-preview.s3.amazonaws.com/fred-wang/trusted-types/pull/557.html" title="Last updated on Oct 23, 2024, 8:57 AM UTC (c072d13)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trusted-types/557/3b74745...fred-wang:c072d13.html" title="Last updated on Oct 23, 2024, 8:57 AM UTC (c072d13)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -676,7 +676,7 @@ Its value is initially « ».
     This function returns the result of the following algorithm:
 
     1. Set |localName| to |tagName| in [=ASCII lowercase=].
-    1. If |elementNs| is an empty string, set |elementNs| to [=HTML namespace=].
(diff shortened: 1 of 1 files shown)


fred-wang:
cc @lukewarlow @koto

Issue #534, opened 2024-07-11
Should "Get Trusted Type compliant string" check `isHTML`/`isScript`/`isScriptURL`?


mbrodesser-Igalia:
https://w3c.github.io/trusted-types/dist/spec/#get-trusted-type-compliant-string-algorithm step 1 currently specifies
"If input has type expectedType". What does that mean? It seems `isHTML` (https://w3c.github.io/trusted-types/dist/spec/#dom-trustedtypepolicyfactory-ishtml) / `isScript` / `isScriptURL` should be invoked.

The callers of "Get Trusted Type compliant string", e.g.`someElement.insertAdjacentHTML` (https://html.spec.whatwg.org/#dom-parsing-and-serialization:dom-element-insertadjacenthtml) don't check that either so it should be checked somewhere.

koto:
https://github.com/w3c/trusted-types/pull/559 should clarify the language, I think. The `is*` functions should not be called there directly, as they can be replaced by user's code, but IIUC the stringification behavior, which is defined for each Trusted Type should correctly extract the value, so the prose looks OK here without having to refer to the object internals?  (see also https://github.com/w3c/trusted-types/issues/541#issuecomment-2449635866).

cc @petervanderbeken @smaug---- if there's a better way to write this up.

mbrodesser-Igalia:
> #559 should clarify the language, I think.

Not sure, see https://github.com/w3c/trusted-types/pull/559#issuecomment-2470588228.

> The `is*` functions should not be called there directly, as they can be replaced by user's code,

Correct.

> but IIUC the stringification behavior, which is defined for each Trusted Type should correctly extract the value,
> so the prose looks OK here without having to refer to the object internals? (see also [#541 (comment)](https://github.com/w3c/trusted-types/issues/541#issuecomment-2449635866)).
> 
> cc @petervanderbeken @smaug---- if there's a better way to write this up.

PR #559, opened 2024-10-31
Replace 'has type' with less confusing 'is an instance of'.

koto:
See https://github.com/w3c/trusted-types/issues/534.

***
<a href="https://pr-preview.s3.amazonaws.com/koto/trusted-types/pull/559.html" title="Last updated on Oct 31, 2024, 11:36 AM UTC (e09da26)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trusted-types/559/e8ff92b...koto:e09da26.html" title="Last updated on Oct 31, 2024, 11:36 AM UTC (e09da26)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -398,7 +398,7 @@ TrustedHTML objects have an associated string <dfn export for="TrustedHTML">data
 The value is set when the object is created, and will never change during its lifetime.
 
 <dfn method for="TrustedHTML">toJSON()</dfn> method steps and the
-<dfn for="TrustedHTML">stringification behavior</dfn> steps of a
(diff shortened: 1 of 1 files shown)


mbrodesser-Igalia:
@koto: is it implied from somewhere to what "instance of" refers to?

If so, from where? Otherwise linking to "instanceof" from [^1] might be the fix. Similar like `parse a JSON string to a JavaScript value` links to the ECMA spec.

[^1]: https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-relational-operators
[^2]: https://infra.spec.whatwg.org/#json

Issue #466, opened 2024-03-05
Creating a policy with policyName="" is possible, but can't be referred to by the "trusted-types" CSP directive


mbrodesser-Igalia:
E.g. https://jsfiddle.net/q5kmL492/ is possible.

https://w3c.github.io/trusted-types/dist/spec/#trusted-types-csp-directive requires the policy-name to consist of at least one character.

That might be annoying when one writes multiple policies named `""` and wants to limit trusted-types to those policies later.

mbrodesser-Igalia:
Adding a keyword `'allow-unnamed'` would fix this.

lukewarlow:
This feels like it shouldn't be allowed? But if we reject unamed policies that might be a compat risk?

mbrodesser-Igalia:
> This feels like it shouldn't be allowed? But if we reject unamed policies that might be a compat risk?

There are use-cases where policy-names are irrelevant. E.g. when allowing all policies via the wildcard `trusted-types *` (https://w3c.github.io/trusted-types/dist/spec/#trusted-types-csp-directive).

bkardell:
I would like to understand if people really do this... Who might have some experience with how common/good an idea (or even just 'why') people would do an unnamed policy? @koto ?

otherdaniel:
> Ww always used a policy name, but they are indeed optional (and only relevant if one guards policy creation by name with `trusted-types` directive).
> 
> @otherdaniel, can we add a use counter for unnamed policies?

Done. (TrustedTypesCreatePolicyWithEmptyName; not sure yet which release it'll appear in.)

koto:
https://chromestatus.com/metrics/feature/timeline/popularity/4897 shows  results in the range of `0.000001` page loads. Just checking with you, @otherdaniel that it's a threshold low enough that we could remove the support for empty policy name?

koto:
Tentatively created https://github.com/w3c/trusted-types/pull/560.

Note that it's still possible to create policies that can not be referred to by CSP, as CSP syntax limits us to https://w3c.github.io/trusted-types/dist/spec/#tt-policy-name. Disallowing creating such policies likely has much bigger backwards compatibility risk though.

fred-wang:
I believe we still don't have tests for that case too.

lukewarlow:
[...]

PR #560, opened 2024-10-31
Disallow empty policy names on creation.

koto:
Fixes https://github.com/w3c/trusted-types/issues/466.


lukewarlow (approved):
Pending compat analysis on this (not sure if a use counter was ever added), this seems fine to me.

PR #562, opened 2024-11-06
Remove link to https://github.com/whatwg/dom/pull/1258

### LINKED ISSUE #537: Does the DOM spec really need to special case script elements for replaceChildren etc.
Currently the [DOM spec](https://dom.spec.whatwg.org/#converting-nodes-into-a-node), and both WebKit and Chromium have special casing in DOM APIs such as node.replaceChildren for script elements. Such that if you try and replace the children with a string or a Text node it would cause a Trusted Types violation.

This was specced in DOM to match the chromium implementation, but was never originally specified in the TT spec.

It's just struck me that this code might be entirely pointless?

Even if you use Trusted types for these APIs they won't set the [script elements internal slot](https://w3c.github.io/trusted-types/dist/spec/#htmlscriptelement-script-text) to actually allow execution. You can also trivially "bypass" them by calling `appendChild(new Text(''))`?

So my question is, why are these implemented in Chromium? Is there something I'm missing or can we drop them from the spec and implementations?

cc @koto @otherdaniel @mbrodesser-Igalia

fred-wang:
This was merged and reverted in
https://github.com/whatwg/dom/commit/809bfa24535b1d55d761d8add353e1c3fde05801

closes #537

***
<a href="https://pr-preview.s3.amazonaws.com/fred-wang/trusted-types/pull/562.html" title="Last updated on Nov 6, 2024, 8:35 AM UTC (f2f903a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trusted-types/562/72ecd2c...fred-wang:f2f903a.html" title="Last updated on Nov 6, 2024, 8:35 AM UTC (f2f903a)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1273,7 +1273,7 @@ Issue: There's no proper definition for the processing of SVG script elements. H
 
 ## Integration with DOM ## {#integration-with-dom}
 
-Note: See [https://github.com/whatwg/dom/pull/1258](https://github.com/whatwg/dom/pull/1258) and [https://github.com/whatwg/dom/pull/1268](https://github.com/whatwg/dom/pull/1268) which upstream this integration.
+Note: See [https://github.com/whatwg/dom/pull/1268](https://github.com/whatwg/dom/pull/1268) which upstreams this integration.
 
 ## Integration with Content Security Policy ## {#integration-with-content-security-policy}
 


fred-wang:
cc @lukewarlow

lukewarlow (approved):

fred-wang:
cc @koto

PR #564, opened 2024-11-15
Put trusted type first when declaring union in IDL files

fred-wang:
This is for consistency with how it is done in the HTML specification and in the PR to DOM (https://github.com/whatwg/dom/pull/1268). This is not web-exposed but may matter for implementations. For example, they may generate different C++ structure to represent `(TrustedHTML or DOMString)` and `(DOMString or TrustedHTML)`.

***
<a href="https://pr-preview.s3.amazonaws.com/fred-wang/trusted-types/pull/564.html" title="Last updated on Nov 18, 2024, 7:27 AM UTC (8e1fa0f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trusted-types/564/4317ed0...fred-wang:8e1fa0f.html" title="Last updated on Nov 18, 2024, 7:27 AM UTC (8e1fa0f)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1146,10 +1146,10 @@ Note: Using these IDL attributes is the recommended way of dynamically setting t
 
 <pre class="idl exclude">
 partial interface HTMLScriptElement {
- [CEReactions] attribute ([LegacyNullToEmptyString] DOMString or TrustedScript) innerText;
- [CEReactions] attribute (DOMString or TrustedScript)? textContent;
- [CEReactions] attribute (USVString or TrustedScriptURL) src;
- [CEReactions] attribute (DOMString or TrustedScript) text;
+ [CEReactions] attribute (TrustedScript or [LegacyNullToEmptyString] DOMString) innerText;
+ [CEReactions] attribute (TrustedScript or DOMString)? textContent;
+ [CEReactions] attribute (TrustedScriptURL or USVString) src;
+ [CEReactions] attribute (TrustedScript or DOMString) text;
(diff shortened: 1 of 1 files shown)


fred-wang:
cc @koto @lukewarlow

spec/index.bs
```diff
@@ -1146,10 +1146,10 @@ Note: Using these IDL attributes is the recommended way of dynamically setting t
 
 <pre class="idl exclude">
 partial interface HTMLScriptElement {
- [CEReactions] attribute ([LegacyNullToEmptyString] DOMString or TrustedScript) innerText;
- [CEReactions] attribute (DOMString or TrustedScript)? textContent;
- [CEReactions] attribute (USVString or TrustedScriptURL) src;
- [CEReactions] attribute (DOMString or TrustedScript) text;
+ [CEReactions] attribute ([LegacyNullToEmptyString] TrustedScript or DOMString) innerText;
```
lukewarlow, line 1149: The idl extended attribute needs moving next to the Dom string type iirc (aka after the or

lukewarlow (changes requested):

spec/index.bs
lukewarlow: Having said that it doesn't appear to be on the original code so I think it might need removing entirely?

fred-wang: For reference, the HTMLElement definition: https://html.spec.whatwg.org/#elements-in-the-dom:dom-innertext

Issue #594, opened 2025-08-07
Streaming support


noamr:
There are new proposals coming for allow streaming HTML into an existing document. See https://github.com/whatwg/html/issues/11542 and https://github.com/whatwg/html/issues/2142#issuecomment-3163420231. 

For this to work, it would need to come along with [sanitation support](https://github.com/WICG/sanitizer-api/issues/190).

To connect the dots, trusted-types would need some form of sanitizing an HTML stream on the go, probably using a `TransformStream`.

This can have the following shape:`policy.createHTMLTransformStream()`, where the source of the untrusted HTML would be piped to the result's `writable` and the `readable` would be piped to the element receiving the markup.

foolip:
I'm trying to make the API surface concrete enough to start writing spec text. The starting point in the [explainer](https://github.com/WICG/declarative-partial-updates/blob/main/patching-explainer.md) is a `patchUnsafe()` method that returns a `WritableStream` objects. Fitting TT into that, all the alternatives I can think of are:

- Let `patchUnsafe()` return a `WritableStream` with replaced internals that always passes chunks through TT. (May need something new in Streams to prevent tampering.)
- Treating the TT step as a `TransformStream` and let `patchUnsafe()` return a `TransformStream` already piped to a `WritableStream`. (May need something new in Stream to prevent them from being unlocked.)
- A new `createHTMLStream()` method on [`TrustedTypePolicy`](https://w3c.github.io/trusted-types/dist/spec/#trusted-type-policy) and let `patchUnsafe()` taking a stream as an optional argument, which it would also return. A stream created from the default policy would be used by default if no argument is given.

For all options there's also the question of whether a plain `WritableStream` or `TransformStream` is OK here, or if a new `TrustedHTMLStream` is needed. Opinions on precise API shape welcome :)

Other than API shape my main question is the same as @noamr had, how to handle dangerous markup being written in small chunks.

lukewarlow:
> If there is a policy without that method, we can either:

- buffer everything and fall back to CreateHTML

- Pass each string separately to createHTML, but l'm not sure that's safe enough.

Alternatively we could also just throw a type error like we would in most other cases.

foolip:
> The way I see it, the TT step should be a `TransformStream` (with a new `createHTMLStream` method or so if exists), and its `writable` part would be what's returned to the caller of `patchUnsafe` by the platform.

[...]

PR #606, opened 2026-02-17
Enable createParserOptions, passing a sanitizer to trusted types

noamr:
`createParserOptions` receives an optional `SetHTMLUnsafeOptions` dictionary (or `ParseHTMLUnsafeOptions`) and turns it into a `TrustedHTMLParserOptions` object using the `createParserOptions` method defined on the policy.

In modern browsers supporting this method, `createParserOptions` serves as the primary Trusted Types mechanism for HTML fragment parsing sinks (`setHTMLUnsafe`, `innerHTML`, `outerHTML`, `insertAdjacentHTML`, and `createContextualFragment`), allowing a policy to configure a `Sanitizer` and/or disallow script execution:
- **Caller-supplied:** Passing a `TrustedHTMLParserOptions` object directly into a fragment parsing sink satisfies Trusted Types enforcement for HTML documents, allowing raw markup strings to be parsed without requiring a `TrustedHTML` object or invoking `createHTML`.
- **Default policy:** When a default policy defines `createParserOptions`, calls to fragment parsing sinks automatically produce a `TrustedHTMLParserOptions` that satisfies the sink, applying the policy's sanitizer and script-execution settings. Note that the default policy takes precedence even if less restrictive than the caller - it can vet a less restrictive sanitizer, or enable script execution.
- **Fallback to `createHTML`:** If no `TrustedHTMLParserOptions` is provided (either by the caller or via a default policy's `createParserOptions`), we fall back to the existing `get trusted type compliant string`, requiring `input` to be a `TrustedHTML` or passing it to the default policy's `createHTML`. This ensures backwards compatibility for policies that only define `createHTML`, and covers string-only sinks that do not support parser options (e.g. `document.write()` and `iframe.srcdoc`). Note that `TrustedHTML` does not receive special handling when `createParserOptions` is present in the policy - in supported browsers/sinks, `createHTML` and `TrustedHTML` are a fallback only.
- **Streaming HTML:** For sinks where the full HTML is not available up front (such as streaming parsing), `get trusted type compliant parser options` is invoked with `throwIfMissing = true`, requiring the caller to supply a `TrustedHTMLParserOptions` or requiring the default policy to define `createParserOptions`.

### Key Design Choices & Semantics:
- **Opaque interface:** `TrustedHTMLParserOptions` does not expose any interface methods or attributes to JavaScript; it acts as an immutable, opaque token whose internal associated configuration and script-execution settings are inspected directly by specification algorithms.
- **XML documents:** For XML fragment parsing (`documentType` is `"xml"`), `TrustedHTMLParserOptions` does not bypass `createHTML`; compliant string input (`TrustedHTML` or `createHTML`) is always required.
- **Security model:** As documented in the Security Considerations, a `createParserOptions` callback that does not configure a sanitizer to remove unsafe markup (e.g. scripts or event handlers) grants a pass-through to parse arbitrary markup without DOM XSS protection. It is the responsibility of the policy author / security auditor to ensure `createParserOptions` enforces adequate sanitization, or that a `TrustedHTMLParserOptions` object is only passed to trusted source (similar to the policy objects themselves).

As a followup, we should evaluate the (existing) combination of `TrustedHTML` with sanitizer. Opened #615 for that.

This is the `trusted-types` part of #594 together with https://github.com/whatwg/html/pull/12583.
Note that this should only be merged when wired into the HTML standard, once the sanitizer is itself upstreamed.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/trusted-types/pull/606.html" title="Last updated on Sep 8, 2026, 12:46 PM UTC (b8ce5b9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/trusted-types/606/d92fbe8...b8ce5b9.html" title="Last updated on Sep 8, 2026, 12:46 PM UTC (b8ce5b9)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -563,6 +563,28 @@ The value is set when the object is created, and will never change during its li
 <dfn for="TrustedScriptURL">stringification behavior</dfn> steps of a
 TrustedScriptURL object are to return the associated [=TrustedScriptURL/data=] value.
 
+### <dfn interface>TrustedParserOptions</dfn> ### {#trused-parser-options}
+
+The TrustedParserOptions interface represents a dictionary that a developer
+can confidently pass into an [=injection sink=] that will parse it as a URL of
+an external script resource.
+These objects are immutable wrappers around a
+{{SetHTMLUnsafeOptions}} dictionary, constructed via a {{TrustedTypePolicy}}'s
+{{TrustedTypePolicy/createParserOptions(options)|createParserOptions}} method.
+
+<pre class="idl">
+
+typedef (Sanitizer or SanitizerConfig or SanitizerPresets) SanitizerInput;
+
+[Exposed=(Window,Worker)]
+interface TrustedParserOptions {
+  readonly attribute SanitizerInput sanitizer;
+};
+</pre>
+
+{{TrustedParserOptions}} objects have an associated {{SanitizerInput}} <dfn export for="TrustedParserOptions">sanitizer</dfn>.
+The value is set when the object is created, and will never change during its lifetime.
+
 ## <dfn>Policies</dfn> ## {#policies-hdr}
 
 Trusted Types can only be created via user-defined
@@ -659,6 +681,32 @@ that it's called only with no attacker-controlled values.
 })();
 </xmp>
 </div>
+
+### Using together with a sanitizer ### {#trusted-types-with-sanitizer}
+
+While trusted types are built in a way that allows custom sanitizers to be used,
+the {{TrustedTypePolicy/createParserOptions(options)|createParserOptions}} also allows for using them as a way to enforce
+using the built-in sanitizer API. [[SANITIZER-API]]
+
+<div class="example" id="trusted-types-with-sanitizer-example">
+<xmp highlight=js>
+(function renderFootnote() {
+  const footnote = await fetch('/footnote.html').then(r => r.text());
+  footNote.setHTMLUnsafe(footnote, trustedTypes.createPolicy('html', {
+    createParserOptions: options => {
+      const sanitizer = new Sanitizer(options.sanitizer || {});
+      sanitizer.removeElement("script");
+      return {
+        sanitizer,
+      };
+    }
+  }).createParserOptions({}));
+})();
+</xmp>
+</div>
+
+Note: `createParserOptions` does not necessarily replace `createHTML`. If both exist in the policy, the `createHTML` function will be called first,
+and the result will be sanitized using the result of the `createParserOptions` function, if that exists.
 
 ### <dfn interface>TrustedTypePolicyFactory</dfn> ### {#trusted-type-policy-factory}
 
@@ -901,6 +949,7 @@ interface TrustedTypePolicy {
   TrustedHTML createHTML(DOMString input, any... arguments);
   TrustedScript createScript(DOMString input, any... arguments);
   TrustedScriptURL createScriptURL(DOMString input, any... arguments);
+  TrustedParserOptions createParserOptions(SetHTMLUnsafeOptions options, any... arguments);
 };
 </pre>
 
@@ -956,6 +1005,21 @@ Each TrustedTypePolicy object has an associated {{TrustedTypePolicyOptions}} <df
       <dd>arguments</dd>
     </dl>
 
+: <dfn>createParserOptions(options)</dfn>
+::  Returns the
+    result of executing the [=Create a Trusted Type=] algorithm, with the
+    following arguments:
+    <dl>
+      <dt>policy</dt>
+      <dd>[=this=] value</dd>
+      <dt>trustedTypeName</dt>
+      <dd>`"TrustedParserOptions"`</dd>
+      <dt>value</dt>
+      <dd>options</dd>
+      <dt>arguments</dt>
+      <dd>arguments</dd>
+    </dl>
+
 </div>
 
 ### <dfn dictionary>TrustedTypePolicyOptions</dfn> ### {#trusted-type-policy-options}
@@ -970,10 +1034,12 @@ dictionary TrustedTypePolicyOptions {
    CreateHTMLCallback createHTML;
    CreateScriptCallback createScript;
    CreateScriptURLCallback createScriptURL;
+   CreateParserOptionsCallback createParserOptions;
 };
 callback CreateHTMLCallback = DOMString? (DOMString input, any... arguments);
 callback CreateScriptCallback = DOMString? (DOMString input, any... arguments);
 callback CreateScriptURLCallback = USVString? (DOMString input, any... arguments);
+callback CreateParserOptionsCallback = SetHTMLUnsafeOptions? (SetHTMLUnsafeOptions options, any... arguments);
 </pre>
 
 ### <dfn data-lt="default-policy-explanation">Default policy</dfn> ### {#default-policy-hdr}
@@ -1077,9 +1143,10 @@ a string (|policyName|), {{TrustedTypePolicyOptions}} dictionary (|options|), an
 1.  Let |policy| be a new {{TrustedTypePolicy}} object.
 1.  Set |policy|'s `name` property value to |policyName|.
 1.  Set |policy|'s [=TrustedTypePolicy/options=] value to «[
-      "createHTML" -> |options|["{{TrustedTypePolicyOptions/createHTML|createHTML}}",
-      "createScript" -> |options|["{{TrustedTypePolicyOptions/createScript|createScript}}",
-      "createScriptURL" -> |options|["{{TrustedTypePolicyOptions/createScriptURL|createScriptURL}}"
+      "createHTML" -> |options|["{{TrustedTypePolicyOptions/createHTML|createHTML}}"],
+      "createScript" -> |options|["{{TrustedTypePolicyOptions/createScript|createScript}}"],
+      "createScriptURL" -> |options|["{{TrustedTypePolicyOptions/createScriptURL|createScriptURL}}"],
+      "createParserOptions" -> |options|["{{TrustedTypePolicyOptions/createParserOptions|createParserOptions}}"]
     ]».
 1.  If the |policyName| is `default`, set the |factory|'s [=TrustedTypePolicyFactory/default policy=] value to |policy|.
 1.  [=set/append|Append=] |policyName| to |factory|'s [=created policy names=].
@@ -1092,6 +1159,10 @@ a string |value| and a list |arguments|, perform the following steps:
 
 1.  Let |policyValue| be the result of executing [=get trusted type policy value=] with the same arguments as this algorithm and additionally true as |throwIfMissing|.
 1.  If the algorithm threw an error, rethrow the error and abort the following steps.
+1. If |trustedTypeName| is "TrustedParserOptions":
+  1. Let |sanitizer| be an empty {{SanitizerConfig}}.
+  1. If |policyValue| is a [=dictionary=] and has a property named "sanitizer", set |sanitizer| to |policyValue|["sanitizer"].
+  1. Return a new interface of {{TrustedTypePolicyOptions}} with its [=TrustedParserOptions/sanitizer=] set to |sanitizer|.
 1.  Let |dataString| be the result of stringifying |policyValue|.
 1.  If |policyValue| is null or undefined, set |dataString| to the empty string.
 1.  Return a new instance of an interface with a type
@@ -1121,6 +1192,10 @@ a string |value|, a list |arguments|, and a boolean |throwIfMissing|, perform th
       <tr>
         <td>"createScriptURL"</td>
         <td>"TrustedScriptURL"</td>
+      </tr>
+      <tr>
+        <td>"createParserOptions"</td>
+        <td>"TrustedParserOptions"</td>
       </tr>
     </table>
 
@@ -1159,6 +1234,28 @@ To <dfn export>get trusted type compliant string</dfn> a {{TrustedType}} type (|
     1. Throw a TypeError and abort further steps.
 1. Assert: |convertedInput| is an instance of |expectedType|.
 1. Return stringified |convertedInput|.
+
+
+## Get Trusted Type compliant parser options ## {#get-trusted-type-compliant-parser-options-algorithm}
+
+To <dfn export>get Trusted Type compliant parser options</dfn> given a [=realm/global object=] (|global|),
+{{TrustedParserOptions}} or a {{SetHTMLOptions}} or a {{SetHTMLUnsafeOptions}} (|input|), a string (|sink|) and a string (|sinkGroup|), run these steps:
+
+1.  If |input| is an instance of {{TrustedParserOptions}}, return |input| and abort these steps.
+1.  Let |requireTrustedTypes| be the result of executing [=Does sink type require trusted types?=] algorithm,
+    passing |global|, |sinkGroup|, and true.
+1.  If |requireTrustedTypes| is `false`, return |input| and abort these steps.
+1.  Let |convertedInput| be the result of executing [=Process value with a default policy=] with the same arguments as this algorithm.
+1.  If the algorithm threw an error, rethrow the error and abort the following steps.
+1.  If |convertedInput| is `null` or `undefined`, execute the following steps:
+    1.  Let |disposition| be the result of executing [=should sink type mismatch violation be blocked by content security policy?=] algorithm,
+        passing |global|, stringified |input| as |source|, |sinkGroup| and |sink|.
+    1.  If |disposition| is `“Allowed”`, return |input| and abort further steps.
+
+        Note: This step assures that the default policy rejection will be reported, but ignored in a report-only mode.
+    1. Throw a TypeError and abort further steps.
+1. Assert: |convertedInput| is an instance of |expectedType|.
+1. Return |convertedInput|.
 
 ## Process value with a default policy ## {#process-value-with-a-default-policy-algorithm}
 
@@ -1207,7 +1304,7 @@ To <dfn export>get trusted type compliant attribute value</dfn> given a string |
     * |attributeNs|
 1. If |attributeData| is null, then:
     1. If |newValue| is a string, return |newValue|.
-    1. <a>Assert</a>: |newValue| is {{TrustedHTML}} or {{TrustedScript}} or {{TrustedScriptURL}}.
+    1. <a>Assert</a>: |newValue| is {{TrustedHTML}} or {{TrustedScript}} or {{TrustedScriptURL}} or {{TrustedParserOptions}}.
     1. Return |value|'s associated data.
 1. Let |expectedType| be the value of the fourth member of |attributeData|.
 1. Let |sink| be the value of the fifth member of |attributeData|.
@@ -1220,7 +1317,7 @@ To <dfn export>get trusted type compliant attribute value</dfn> given a string |
 
    If the algorithm threw an error, rethrow the error.
 
-## Get Trusted Type data for attribute ## {#get-trusted-type-data-for-attribute}
+## Get Trusted Type data for attribute ## {#get-trusted-type-data-for-attribute-algorithm}
 
 To <dfn>get Trusted Type data for attribute</dfn> given |element|, |attribute|, |attributeNs|, perform the following steps:
 
@@ -1249,7 +1346,7 @@ Issue: The [=event handler content attribute=] concept used below is ambiguous. 
 # Integrations # {#integrations}
 
 <pre class="idl">
-typedef (TrustedHTML or TrustedScript or TrustedScriptURL) TrustedType;
+typedef (TrustedHTML or TrustedScript or TrustedScriptURL or TrustedParserOptions) TrustedType;
 </pre>
 
 ## Integration with HTML ## {#integration-with-html}
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -576,7 +576,7 @@ These objects are immutable wrappers around a
 
 typedef (Sanitizer or SanitizerConfig or SanitizerPresets) SanitizerInput;
 
-[Exposed=(Window,Worker)]
+[Exposed=(Window)]
 interface TrustedParserOptions {
   readonly attribute SanitizerInput sanitizer;
 };
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -692,7 +692,7 @@ using the built-in sanitizer API. [[SANITIZER-API]]
 <xmp highlight=js>
 (function renderFootnote() {
   const footnote = await fetch('/footnote.html').then(r => r.text());
-  footNote.setHTMLUnsafe(footnote, trustedTypes.createPolicy('html', {
+  const policy = trustedTypes.createPolicy('html', {
     createParserOptions: options => {
       const sanitizer = new Sanitizer(options.sanitizer || {});
       sanitizer.removeElement("script");
@@ -700,7 +700,8 @@ using the built-in sanitizer API. [[SANITIZER-API]]
         sanitizer,
       };
     }
-  }).createParserOptions({}));
+  });
+  footNote.setHTMLUnsafe(footnote, policy.createParserOptions({}));
 })();
 </xmp>
 </div>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1136,6 +1136,73 @@ via using the <a http-header>Content-Security-Policy-Report-Only</a> HTTP Respon
 Note: Most of the enforcement rules are defined as modifications of the
 algorithms in other specifications, see [[#integrations]].
 
+### Sanitization enforcement with {{TrustedTypePolicy/createParserOptions(options)|createParserOptions}} ### {#sanitization-enforcement-hdr}
+
+While {{TrustedTypePolicyOptions/createHTML|createHTML}} relies on sanitizing strings before they are injected to the parser using a JS sanitizer,
+the {{TrustedTypePolicy/createParserOptions(options)|createParserOptions}} allows the page's default policy to enforce HTML sanitization.
+
+A {{TrustedTypePolicy/createParserOptions(options)|createParserOptions}} callback in a required default policy would run for
+any attempt to inject HTML into a document using script.
+
+<div class="example" id="require-sanitizer-for-html-injection">
+Enforce Trusted Types at the DOM XSS injection sinks.
+
+<pre class="http">
+Content-Security-Policy: require-trusted-types-for 'script'
+</pre>
+
+Inject a sanitizer when using an HTML injection sink.
+
+<xmp highlight=js>
+(function renderFootnote() {
+  const defaultPolicy = trustedTypes.createPolicy('default', {
+    createParserOptions: options => {
+      const sanitizer = new Sanitizer(options.sanitizer)
+      sanitizer.removeUnsafe();
+      return { ...options, sanitizer };
+    }
+  });
+
+  const trustedOptions = trustedTypes.createPolicy('trusted', {
+    createParserOptions: options => options
+  }).createParserOptions({});
+  const untrustedFootnoteContent = await fetch('/footnote.html').then(r => r.text());
+
+  // This HTML will be sanitized with a "safe" HTML sanitizer, as the default policy injects
+  // a sanitizer with unsafe markup removed.
+  footNote.innerHTML = untrustedFootnoteContent;
+  // or:
+  footNote.setHTMLUnsafe(untrustedFootnoteContent, {});
+  const trustedMainContent = await fetch("main.html").then(r => r.text());
+  main.setHTMLUnsafe(trustedMainContent, trustedOptions);
+})();
+</xmp>
+
+Apart from sanizitation, the default policy can also prevent an HTML injection from running scripts.
+
+<xmp highlight=js>
+(function renderFootnote() {
+  const defaultPolicy = trustedTypes.createPolicy('default', {
+    createParserOptions: options => {
+      return { ...options, runScripts: false };
+    }
+  });
+
+  // The default policy would override the {{SetHTMLUnsafeOptions/runScripts}} value to false.
+  main.setHTMLUnsafe(someHTMLWithScripts, {runScripts: true});
+
+  // This applies to {{Range/createContextualFragment(string)}}
+  const fragment = new Range().createContextualFragment("<script>alert(1)</" + "script>");
+  const div = document.createElement('div');
+  // The scripts in the fragment would not be executed, as the default policy overrides {{SetHTMLUnsafeOptions/runScripts}}.
+  div.appendChild(fragment);
+})();
+</xmp>
+
+</div>
+
+
+
 # Algorithms # {#algorithms}
 
 ## Create a Trusted Type Policy ## {#create-trusted-type-policy-algorithm}
@@ -1225,7 +1292,9 @@ a string |value|, a list |arguments|, and a boolean |throwIfMissing|, perform th
 ## Get Trusted Type parser input ## {#get-trusted-type-parser-input-algorithm}
 
 To <dfn export>get Trusted Type compliant input</dfn> given a {{TrustedType}} |expectedType|, a [=realm/global object=] (|global|), string, or <code data-x="">Stream</code> |input|,
-a {{TrustedParserOptions}}, {{SetHTMLUnsafeOptions}} or <code data-x="">Legacy</code> |options|, a string (|sink|) and a string (|sinkGroup|), run these steps:
+a {{TrustedParserOptions}}, {{SetHTMLUnsafeOptions}} or <code data-x="">Legacy</code> |options|, and a string (|sink|), run these steps:
+
+1. Let |sinkGroup| be 
 
 1. If |input| is not a <code data-x="">Stream</code>:
   1. Set |input| to the result of calling get trusted type compliant string, passing |expectedType|, |global|, |input|, |sink| and |sinkGroup|.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1523,7 +1523,7 @@ Issue: The [=event handler content attribute=] concept used below is ambiguous. 
 # Integrations # {#integrations}
 
 <pre class="idl">
-typedef (TrustedHTML or TrustedScript or TrustedScriptURL or TrustedParserOptions) TrustedType;
+typedef (TrustedHTML or TrustedScript or TrustedScriptURL) TrustedType;
 </pre>
 
 ## Integration with HTML ## {#integration-with-html}
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -573,12 +573,11 @@ TrustedScriptURL object are to return the associated [=TrustedScriptURL/data=] v
 
 ### <dfn interface>TrustedParserOptions</dfn> ### {#trused-parser-options}
 
-The {{TrustedParserOptions}} interface represents a dictionary that a developer
-can confidently pass into an [=injection sink=] that will parse it as a URL of
-an external script resource.
+The {{TrustedParserOptions}} interface represents options that a developer
+can confidently pass into an [=injection sink=] that will parse HTML.
 These objects are immutable wrappers around a
 {{SetHTMLUnsafeOptions}} dictionary, constructed via a {{TrustedTypePolicy}}'s
-{{TrustedTypePolicy/createParserOptions(options)|createParserOptions}} method.
+{{TrustedTypePolicy/createParserOptions(options, ...arguments)|createParserOptions}} method.
 
 <pre class="idl">
 
@@ -1032,7 +1031,7 @@ Each TrustedTypePolicy object has an associated {{TrustedTypePolicyOptions}} <df
       <dd>arguments</dd>
     </dl>
 
-: <dfn>createParserOptions(options)</dfn>
+: <dfn>createParserOptions(options, ...arguments)</dfn>
 ::  Returns the
     result of executing the [=Create a Trusted Type=] algorithm, with the
     following arguments:
@@ -1195,7 +1194,7 @@ Inject a sanitizer when using an HTML injection sink.
 })();
 </xmp>
 
-Apart from sanizitation, the default policy can also prevent an HTML injection from running scripts.
+Apart from sanitization, the default policy can also prevent an HTML injection from running scripts.
 HTML inserted with {{SetHTMLUnsafeOptions/runScripts}} set to false can still insert `script` elements into the document,
 but those scripts are not automatically executed.
 
@@ -1251,19 +1250,19 @@ a string (|policyName|), {{TrustedTypePolicyOptions}} dictionary (|options|), an
 ## Create a Trusted Type ## {#create-a-trusted-type-algorithm}
 
 To <dfn>create a trusted type</dfn> given a {{TrustedTypePolicy}} |policy|, a type name |trustedTypeName|,
-a string |value| and a list |arguments|, perform the following steps:
+a string or dictionary |value| and a list |arguments|, perform the following steps:
 
 1.  Let |policyValue| be the result of executing [=get trusted type policy value=] with the same arguments as this algorithm and additionally true as |throwIfMissing|.
 1.  If the algorithm threw an error, rethrow the error and abort the following steps.
 1.  If |trustedTypeName| is `"TrustedParserOptions"`:
-    1.  Let |sanitizer| be a new {{SanitizerConfig}}.
+    1.  Let |sanitizer| be null.
     1.  Let |runScripts| be false.
     1.  If |policyValue| is a [=dictionary=]:
         1.  Set |sanitizer| to the result of
             [=get a sanitizer config from dictionary|getting a sanitizer configuration=]
             from |policyValue|.
         1.  Set |runScripts| to |policyValue|["runScripts"] [=with default=] false.
-    1.  Return a new instance of {{TrustedParserOptions}} with its
+    1.  Return a new instance of {{TrustedParserOptions}} in |policy|'s [=relevant realm=] with its
         [=TrustedParserOptions/sanitizer config=] set to |sanitizer|, and its
         [=TrustedParserOptions/runScripts=] set to |runScripts|.
 1.  Let |dataString| be the result of stringifying |policyValue|.
@@ -1274,7 +1273,7 @@ a string |value| and a list |arguments|, perform the following steps:
 ## Get Trusted Type policy value ## {#get-trusted-type-policy-value-algorithm}
 
 To <dfn>get trusted type policy value</dfn> given a {{TrustedTypePolicy}} |policy|, a type name |trustedTypeName|,
-a string |value|, a list |arguments|, and a boolean |throwIfMissing|, perform the following steps:
+a string or dictionary |value|, a list |arguments|, and a boolean |throwIfMissing|, perform the following steps:
 
 1.  Let |functionName| be a function name for the given |trustedTypeName|,
     based on the following table:
@@ -1341,7 +1340,7 @@ To <dfn export>get trusted type compliant string</dfn> a {{TrustedType}} type (|
 1.  If the algorithm threw an error, rethrow the error and abort the following steps.
 1.  If |convertedInput| is `null` or `undefined`, execute the following steps:
     1.  Let |disposition| be the result of executing [=should sink type mismatch violation be blocked by content security policy?=] algorithm,
-        passing |global|, stringified |input| as |source|, |sinkGroup| and |sink|.
+        passing |global|, |sink|, |sinkGroup|, and stringified |input| as |source|.
     1.  If |disposition| is `“Allowed”`, return stringified |input| and abort further steps.
 
         Note: This step assures that the default policy rejection will be reported, but ignored in a report-only mode.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -660,6 +682,32 @@ that it's called only with no attacker-controlled values.
...
+
+While trusted types are built in a way that allows custom sanitizers to be used,
+the {{TrustedTypePolicy/createParserOptions(options)|createParserOptions}} also allows for using them as a way to enforce
+using the built-in sanitizer API. [[SANITIZER-API]]
+
+<div class="example" id="trusted-types-with-sanitizer-example">
+<xmp highlight=js>
+(function renderFootnote() {
+  const footnote = await fetch('/footnote.html').then(r => r.text());
+  footNote.setHTMLUnsafe(footnote, trustedTypes.createPolicy('html', {
```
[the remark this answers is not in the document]
noamr: Done

noamr:
I've made some changes:
- Some of the algorithm calls from "get Trusted Type compliant parser options" assumed stringifications. Made that algorithm do things on its own and passing a new object (repersented as an ordered map) instead.
- Allow passing `throwIfMissing` from the outside. `streamHTMLUnsafe` methods should throw if the policy cannot create parser options, while `setHTMLUnsafe` and friends can rely on `createHTML`.
- Support `runScripts` as per https://github.com/whatwg/html/pull/12495

(Note: this PR will fail build until we merge the latter)

lukewarlow:
Are you able to fix the build failure now?

noamr:
> Are you able to fix the build failure now?

I'll revisit this once I have the HTML counterpart ready.

spec/index.bs
```diff
@@ -1134,6 +1222,20 @@ a string |value|, a list |arguments|, and a boolean |throwIfMissing|, perform th
     |args| and `"rethrow"`.
 1.  Return |policyValue|.
 
+## Get Trusted Type parser input ## {#get-trusted-type-parser-input-algorithm}
+
+To <dfn export>get Trusted Type compliant input</dfn> given a {{TrustedType}} |expectedType|, a [=realm/global object=] (|global|), string, or <code data-x="">Stream</code> |input|,
```
annevk: What is `Stream`?

spec/index.bs
```diff
@@ -1134,6 +1222,20 @@ a string |value|, a list |arguments|, and a boolean |throwIfMissing|, perform th
...
+a {{TrustedParserOptions}}, {{SetHTMLUnsafeOptions}} or <code data-x="">Legacy</code> |options|, a string (|sink|) and a string (|sinkGroup|), run these steps:
```
annevk:
```suggestion
a {{TrustedParserOptions}}, {{SetHTMLUnsafeOptions}} or <code data-x="">Legacy</code> |options|, a string (|sink|) and a string (|sinkGroup|):
```
Also why are the variable names in parentheses?

spec/index.bs
```diff
@@ -1134,6 +1222,20 @@ a string |value|, a list |arguments|, and a boolean |throwIfMissing|, perform th
...
+
+1. If |input| is not a <code data-x="">Stream</code>:
+  1. Set |input| to the result of calling get trusted type compliant string, passing |expectedType|, |global|, |input|, |sink| and |sinkGroup|.
```
annevk: Missing Oxford comma.

spec/index.bs
```diff
@@ -1134,6 +1222,20 @@ a string |value|, a list |arguments|, and a boolean |throwIfMissing|, perform th
...
+  1. If the algorithm threw an error, rethrow the error and abort the following steps.
+1. If |options| is <code data-x="">Legacy</code>, set |options| to a new {{SetHTMLUnsafeOptions}} dictionary.
+1. Let |throwIfMissing| be true if |input| is not a <code data-x="">Stream</code>; Otherwise false.
```
annevk: otherwise should not start with uppercase here

spec/index.bs
```diff
@@ -1160,6 +1262,46 @@ To <dfn export>get trusted type compliant string</dfn> a {{TrustedType}} type (|
...
+1.  Let |function| be |defaultPolicy|'s [=TrustedTypePolicy/options=]["createParserOptions"].
+1.  If |function| is null:
+    1. If |throwIfMissing| is true, throw a {{TypeError}}.
+    1. Return |input|.
+1. Let |argument| be an [=/ordered map=] whose [=map/entries=] represent the [=dictionary=] passed as |input|.
+
+  Note: This ensures the policy cannot mutate the original options.
+
+1. Let |policyValue| be the result of [=invoke|invoking=] |function| with « |argument| » and `"rethrow"`.
+1. If the algorithm threw an error, rethrow the error and abort the following steps.
```
annevk, line 1395: Throw/rethrow already terminates the current steps so better not state it in duplicate to avoid confusion.

spec/index.bs
```diff
@@ -1160,6 +1262,46 @@ To <dfn export>get trusted type compliant string</dfn> a {{TrustedType}} type (|
...
+1. If |policyValue| is `null` or `undefined`, execute the following steps:
```
annevk:
```suggestion
1. If |policyValue| is `null` or `undefined`:
```

spec/index.bs
```diff
@@ -1160,6 +1262,46 @@ To <dfn export>get trusted type compliant string</dfn> a {{TrustedType}} type (|
...
+    1. Let |disposition| be the result of executing [=should sink type mismatch violation be blocked by content security policy?=] algorithm,
+        passing |global|, |input|, |sinkGroup| and |sink|.
+    1. If |disposition| is `“Allowed”`, return |input| and abort further steps.
+
+        Note: This step assures that the default policy rejection will be reported, but ignored in a report-only mode.
+
+    1. Throw a TypeError and abort further steps.
```
annevk:
```suggestion
    1. Throw a {{TypeError}}.
```

spec/index.bs
```diff
@@ -1160,6 +1262,46 @@ To <dfn export>get trusted type compliant string</dfn> a {{TrustedType}} type (|
...
+1. Let |argument| be an [=/ordered map=] whose [=map/entries=] represent the [=dictionary=] passed as |input|.
+
+  Note: This ensures the policy cannot mutate the original options.
+
+1. Let |policyValue| be the result of [=invoke|invoking=] |function| with « |argument| » and `"rethrow"`.
+1. If the algorithm threw an error, rethrow the error and abort the following steps.
+1. If |policyValue| is `null` or `undefined`, execute the following steps:
+    1. Let |disposition| be the result of executing [=should sink type mismatch violation be blocked by content security policy?=] algorithm,
+        passing |global|, |input|, |sinkGroup| and |sink|.
+    1. If |disposition| is `“Allowed”`, return |input| and abort further steps.
```
annevk:
```suggestion
    1. If |disposition| is `“Allowed”`, return |input|.
```

spec/index.bs
```diff
@@ -1160,6 +1262,46 @@ To <dfn export>get trusted type compliant string</dfn> a {{TrustedType}} type (|
...
+
+        Note: This step assures that the default policy rejection will be reported, but ignored in a report-only mode.
+
+    1. Throw a TypeError and abort further steps.
+1. Let |sanitizer| be |policyValue|["`sanitizer`"], [=with default|defaulting to=] a new {{SanitizerConfig}}.
+1. If |sanitizer| is a {{Sanitizer}}, then set |sanitizer| to a new {{Sanitizer}} whose [=Sanitizer/configuration=] is the result of [=canonicali`ze the configuration|canonicalizing=] |sanitizer|'s [=Sanitizer/configuration=].
+
+  This step assures that the provided {{Sanitizer}} is not mutated by the policy.
```
annevk: Is this supposed to be a note?

spec/index.bs
```diff
@@ -1160,6 +1262,46 @@ To <dfn export>get trusted type compliant string</dfn> a {{TrustedType}} type (|
...
+1. If |policyValue| is `null` or `undefined`, execute the following steps:
+    1. Let |disposition| be the result of executing [=should sink type mismatch violation be blocked by content security policy?=] algorithm,
+        passing |global|, |input|, |sinkGroup| and |sink|.
+    1. If |disposition| is `“Allowed”`, return |input| and abort further steps.
+
+        Note: This step assures that the default policy rejection will be reported, but ignored in a report-only mode.
+
+    1. Throw a TypeError and abort further steps.
+1. Let |sanitizer| be |policyValue|["`sanitizer`"], [=with default|defaulting to=] a new {{SanitizerConfig}}.
+1. If |sanitizer| is a {{Sanitizer}}, then set |sanitizer| to a new {{Sanitizer}} whose [=Sanitizer/configuration=] is the result of [=canonicali`ze the configuration|canonicalizing=] |sanitizer|'s [=Sanitizer/configuration=].
```
annevk: Shouldn't that already be canonicalized?

annevk:
Also, reportedly this impacts `innerHTML` and friends as well somehow? I think that warrants some examples.

spec/index.bs
noamr: It's a special value meaning that we don't have a string yet and we are going to stream the HTML. is null better? This seemed more descriptive.

noamr:
> Also, reportedly this impacts `innerHTML` and friends as well somehow? I think that warrants some examples.

Added prose and examples for this

spec/index.bs
noamr: This is the equivalent of cloning, to avoid mutating an existing object. The note underneath says so

spec/index.bs
noamr: Done

spec/index.bs
noamr: Done

spec/index.bs
noamr: I think it's copy/pasta. Fixed

noamr:
(The build failure is in main, trying to resolve separately. Do you have an idea maybe @lukewarlow ?)

evilpie:
In `TrustedParserOptions` `sanitizer` member is readonly, but a Sanitizer instance allows for internal modification. Isn't that unexpected?

Also I feel like we should probably always return a Sanitizer object, instead of one of three options? Otherwise it seems easy to accidentally write code (not sure what for though?)  that only works with a subset.

noamr:
> In `TrustedParserOptions` `sanitizer` member is readonly, but a Sanitizer instance allows for internal modification. Isn't that unexpected?

It's a good point. I guess we need to return a clone of the underlying sanitizer if the getter is invoked so that you couldn't get a `TrustedParserOptions` and then mutate it.

> Also I feel like we should probably always return a Sanitizer object, instead of one of three options? Otherwise it seems easy to accidentally write code (not sure what for though?) that only works with a subset.

Yea I was contemplating that. SGTM

spec/index.bs
noamr: Refactored this, doesn't use "canonicalize" anymore.

noamr:
> > In `TrustedParserOptions` `sanitizer` member is readonly, but a Sanitizer instance allows for internal modification. Isn't that unexpected?
> 
> It's a good point. I guess we need to return a clone of the underlying sanitizer if the getter is invoked so that you couldn't get a `TrustedParserOptions` and then mutate it.
> 
> > Also I feel like we should probably always return a Sanitizer object, instead of one of three options? Otherwise it seems easy to accidentally write code (not sure what for though?) that only works with a subset.
> 
> Yea I was contemplating that. SGTM

The new version only passes a `Sanitizer` option around, and mutating a `trustedParserOptions.sanitizer` doesn't affect the underlying sanitizer.

lukewarlow:
Might be a silly question but why does the TrustedParserOptions need to expose the value of sanitizer and run scripts? Can it not be an opaque container than only web APIs can reach into? That way there's no weirdness about mutating a value that doesn't affect anything?

noamr:
> Might be a silly question but why does the TrustedParserOptions need to expose the value of sanitizer and run scripts? Can it not be an opaque container than only web APIs can reach into? That way there's no weirdness about mutating a value that doesn't affect anything?

I guess it's kind of expected as a reflection, the same way you can stringify a `TrustedHTML`?

The mutability thing is a bit awkward but in most cases it wouldn't have any effect.
I'm also open to changing this reflection behavior, it's not material to the design.

lukewarlow:
I have a feeling that a function is a better way to do the access for the sanitizer object. It feels more logical that a function returns a clone of an object rather than directly allowing mutations. But I'm curious for others thoughts here.

noamr:
> I have a feeling that a function is a better way to do the access for the sanitizer object. It feels more logical that a function returns a clone of an object rather than directly allowing mutations. But I'm curious for others thoughts here.

you mean something like `sanitizer()` that returns a fresh `Sanitizer` object in each call?
Yea that also works for me.

noamr:
> > I have a feeling that a function is a better way to do the access for the sanitizer object. It feels more logical that a function returns a clone of an object rather than directly allowing mutations. But I'm curious for others thoughts here.
> 
> you mean something like `sanitizer()` that returns a fresh `Sanitizer` object in each call? Yea that also works for me.

Done

noamr:
@lukewarlow having checked this, I think I have slight preference to keeping it as a getter, as it has the nice ergonomic benefit of being able to pass it to `setHTML` (and the other "safe" variant) as if it was a `SetHTMLOptions` - it reads the  `sanitizer` property from the getter. If it's a method, it fails silently as it doesn't recognize that method as a sanitizer.

lukewarlow:
Can we change it so that just works though? It feels like we could. I'm happy to leave it as a getter if that's preferred though.

noamr:
> Can we change it so that just works though? It feels like we could. I'm happy to leave it as a getter if that's preferred though.

Yea I guess its nicer as a method, and we can allow TrustedParserOptions in the IDL explicitly

spec/index.bs
```diff
@@ -1060,6 +1148,73 @@ via using the <a http-header>Content-Security-Policy-Report-Only</a> HTTP Respon
...
+  // a sanitizer with unsafe markup removed.
+  footNote.innerHTML = untrustedFootnoteContent;
+  // or:
+  footNote.setHTMLUnsafe(untrustedFootnoteContent, {});
+  const trustedMainContent = await fetch("main.html").then(r => r.text());
+  main.setHTMLUnsafe(trustedMainContent, trustedOptions);
+})();
+</xmp>
+
+Apart from sanizitation, the default policy can also prevent an HTML injection from running scripts.
```
lukewarlow: Might be worth explaining what the difference is between preventing running scripts and then the sanitizing of scripts?

Though as this is a niche spec perhaps no Devs are actually reading it.

spec/index.bs
```diff
@@ -1092,6 +1248,17 @@ a string |value| and a list |arguments|, perform the following steps:
 
 1.  Let |policyValue| be the result of executing [=get trusted type policy value=] with the same arguments as this algorithm and additionally true as |throwIfMissing|.
 1.  If the algorithm threw an error, rethrow the error and abort the following steps.
+1. If |trustedTypeName| is "TrustedParserOptions":
+  1. Let |sanitizer| be a new {{SanitizerConfig}}.
```
lukewarlow: The nesting here isn't quite right it's not seeing this as a sublist.

spec/index.bs
```diff
@@ -1134,6 +1305,22 @@ a string |value|, a list |arguments|, and a boolean |throwIfMissing|, perform th
     |args| and `"rethrow"`.
 1.  Return |policyValue|.
 
+## Get Trusted Type parser input ## {#get-trusted-type-parser-input-algorithm}
+
+To <dfn export>get Trusted Type compliant input</dfn> given a {{TrustedType}} |expectedType|, a [=realm/global object=] |global|, a string or <code data-x="">Stream</code> |input|,
+a {{TrustedParserOptions}}, {{SetHTMLUnsafeOptions}} or <code data-x="">Legacy</code> |options|, and a string |sink|:
+
+1. Let |sinkGroup| be
```
lukewarlow: This is missing a value?

spec/index.bs
```diff
@@ -1134,6 +1305,22 @@ a string |value|, a list |arguments|, and a boolean |throwIfMissing|, perform th
...
 
+## Get Trusted Type parser input ## {#get-trusted-type-parser-input-algorithm}
+
+To <dfn export>get Trusted Type compliant input</dfn> given a {{TrustedType}} |expectedType|, a [=realm/global object=] |global|, a string or <code data-x="">Stream</code> |input|,
+a {{TrustedParserOptions}}, {{SetHTMLUnsafeOptions}} or <code data-x="">Legacy</code> |options|, and a string |sink|:
+
+1. Let |sinkGroup| be 
+
+1. If |input| is not a <code data-x="">Stream</code>:
+  1. Set |input| to the result of calling get trusted type compliant string, passing |expectedType|, |global|, |input|, |sink|, and |sinkGroup|.
```
lukewarlow: This is also not nested quite right and so isn't being treated as a sublist of steps.

lukewarlow:
Other than the 3 changes above this PR LGTM. Haven't reviewed the HTML side yet.

spec/index.bs
noamr: Fixed the whole `sinkGroup` flow

spec/index.bs
noamr: Added some words about this

spec/index.bs
```diff
@@ -1160,6 +1349,51 @@ To <dfn export>get trusted type compliant string</dfn> a {{TrustedType}} type (|
...
+1. If |policyValue| is `null` or `undefined`:
+    1. Let |disposition| be the result of executing [=should sink type mismatch violation be blocked by content security policy?=] algorithm,
+        passing |global|, |input|, `"TrustedHTML"` and |sink|.
+    1. If |disposition| is `“Allowed”`, return |input|.
+
+        Note: This step assures that the default policy rejection will be reported, but ignored in a report-only mode.
+
+    1. Throw a {{TypeError}}.
+1. Let |runScripts| be true if |policyValue|["`runScripts`"] is true and |input| is a {{SetHTMLUnsafeOptions}} whose {{SetHTMLUnsafeOptions/runScripts}} is true.
+1. Return a new {{TrustedParserOptions}} in |input|'s [=relevant realm=]
```
zcorpan: *input* is a dictionary which is not a platform object, so relevant realm is not defined. Should this use *global*'s relevant realm?

spec/index.bs
```diff
@@ -1134,6 +1312,17 @@ a string |value|, a list |arguments|, and a boolean |throwIfMissing|, perform th
     |args| and `"rethrow"`.
 1.  Return |policyValue|.
 
+## Get Trusted Type parser input ## {#get-trusted-type-parser-input-algorithm}
+
+To <dfn export>get Trusted Type compliant input</dfn> given a [=realm/global object=] |global|, a string |input|,
+a {{TrustedParserOptions}} or {{SetHTMLUnsafeOptions}} |options|, and a string |sink|:
```
zcorpan: Missing `ParseHTMLUnsafeOptions`

spec/index.bs
```diff
@@ -1160,6 +1349,51 @@ To <dfn export>get trusted type compliant string</dfn> a {{TrustedType}} type (|
...
+1.  If |requireTrustedTypes| is `false`, return |input|.
+1.  Let |defaultPolicy| be the value of |global|'s [=Window/trusted type policy factory=]'s [=TrustedTypePolicyFactory/default policy=].
+1.  If |defaultPolicy| is null, throw a {{TypeError}}.
+1.  Let |function| be |defaultPolicy|'s [=TrustedTypePolicy/options=]["createParserOptions"].
+1.  If |function| is null:
+    1. If |throwIfMissing| is true, throw a {{TypeError}}.
+    1. Return |input|.
+1. Let |runScriptsInput| be |input|["runScripts"], [=with default=] false.
+1. Let |sanitizerInputConfig| be the results of [=get a sanitizer config from dictionary|getting a sanitizer config=] from |input|.
+1. Let |sanitizerInput| be null if |sanitizerInputConfig| is null; otherwise a new {{Sanitizer}}, [=configure|configured=] withh |sanitizerInputConfig|.
```
zcorpan: Typo
```suggestion
1. Let |sanitizerInput| be null if |sanitizerInputConfig| is null; otherwise a new {{Sanitizer}}, [=configure|configured=] with |sanitizerInputConfig|.
```

spec/index.bs
```diff
@@ -563,6 +571,42 @@ The value is set when the object is created, and will never change during its li
 <dfn for="TrustedScriptURL">stringification behavior</dfn> steps of a
 TrustedScriptURL object are to return the associated [=TrustedScriptURL/data=] value.
 
+### <dfn interface>TrustedParserOptions</dfn> ### {#trused-parser-options}
+
+The TrustedParserOptions interface represents a dictionary that a developer
```
zcorpan:
```suggestion
The {{TrustedParserOptions}} interface represents a dictionary that a developer
```

zcorpan (approved):

zcorpan:
I let chatgpt re-review:

<h2>1. Blocking: parser-options enforcement checks a nonexistent CSP sink group</h2>
<p>In “get Trusted Type compliant parser options”, the CSP check uses <code>"TrustedHTML"</code> as the sink group. The defined <code>require-trusted-types-for</code> sink group is <code>"script"</code>, and the sibling parser-input path already uses <code>"script"</code>. As written, <code>require-trusted-types-for 'script'</code> will not match, so <code>createParserOptions</code> can be skipped. Please use <code>"script"</code> in both enforcement and violation-reporting paths and add enforced/report-only CSP coverage.</p>
<h2>2. Blocking: the CSP violation-reporting call has the wrong arguments</h2>
<p>The parser-options violation-reporting call does not match the CSP algorithm’s <code>(global, sink, sinkGroup, source)</code> signature. It currently passes the options input in the <code>sink</code> slot and the sink name in the <code>source</code> slot, in addition to using the wrong sink group. Please pass the real sink, <code>"script"</code>, and a defined string source/sample for the parser-options violation.</p>
<h2>3. Blocking: <code>createParserOptions()</code> passes a dictionary through string-typed algorithms</h2>
<p><code>TrustedTypePolicy.createParserOptions(options)</code> passes a <code>SetHTMLUnsafeOptions</code> dictionary as the <code>value</code> argument to “create a Trusted Type”, but that algorithm and “get trusted type policy value” declare <code>value</code> as a string. Please generalize the value type for the <code>TrustedParserOptions</code> case or use a dedicated parser-options creation algorithm.</p>
<p>There is also nonblocking copy/paste prose describing <code>TrustedParserOptions</code> as the URL of an external script resource.</p>

annevk:
It's not clear to me if this PR is ready or not. It seems from the AI feedback above it's not?

I think this could also do with a much clearer description of what it is adding. Turning something into `TrustedParserOptions` does not really explain how this enables `innerHTML` to now perform sanitization, for instance.

noamr:
> I let chatgpt re-review:
> 
> ## 1. Blocking: parser-options enforcement checks a nonexistent CSP sink group
> In “get Trusted Type compliant parser options”, the CSP check uses `"TrustedHTML"` as the sink group. The defined `require-trusted-types-for` sink group is `"script"`, and the sibling parser-input path already uses `"script"`. As written, `require-trusted-types-for 'script'` will not match, so `createParserOptions` can be skipped. Please use `"script"` in both enforcement and violation-reporting paths and add enforced/report-only CSP coverage.

Fixed

> 
> ## 2. Blocking: the CSP violation-reporting call has the wrong arguments
> The parser-options violation-reporting call does not match the CSP algorithm’s `(global, sink, sinkGroup, source)` signature. It currently passes the options input in the `sink` slot and the sink name in the `source` slot, in addition to using the wrong sink group. Please pass the real sink, `"script"`, and a defined string source/sample for the parser-options violation.

Fixed

> 
> ## 3. Blocking: `createParserOptions()` passes a dictionary through string-typed algorithms
> `TrustedTypePolicy.createParserOptions(options)` passes a `SetHTMLUnsafeOptions` dictionary as the `value` argument to “create a Trusted Type”, but that algorithm and “get trusted type policy value” declare `value` as a string. Please generalize the value type for the `TrustedParserOptions` case or use a dedicated parser-options creation algorithm.

Fixed

> 
> There is also nonblocking copy/paste prose describing `TrustedParserOptions` as the URL of an external script resource.

Fixed

noamr:
> It's not clear to me if this PR is ready or not. It seems from the AI feedback above it's not?
> 
> I think this could also do with a much clearer description of what it is adding. Turning something into `TrustedParserOptions` does not really explain how this enables `innerHTML` to now perform sanitization, for instance.

Fixed those issues and also the commit message.

spec/index.bs
```diff
@@ -1134,6 +1315,17 @@ a string |value|, a list |arguments|, and a boolean |throwIfMissing|, perform th
     |args| and `"rethrow"`.
 1.  Return |policyValue|.
 
+## Get Trusted Type compliant input ## {#get-trusted-type-compliant-input-algorithm}
+
+To <dfn export>get trusted type compliant input</dfn> given a [=realm/global object=] |global|, a string |input|,
```
zcorpan: Should be `a string or {{TrustedHTML}} |input|`

noamr: Done

zcorpan (approved):

noamr:
Revised based on @mozfreddyb's [comment](https://github.com/whatwg/html/pull/12583#issuecomment-5438071714):
we now call `createParserOptions` first, and only call `createHTML` if at that point we don't have a `TrustedParserOptions`.

This allows policy authors to prefer the platform sanitizer, and to fall back to `createHTML` for non-supporting browsers or for `document.write()` / `srcdoc`.

spec/index.bs
```diff
@@ -1152,14 +1344,76 @@ To <dfn export>get trusted type compliant string</dfn> a {{TrustedType}} type (|
...
+        1. Let |disposition| be the result of executing [=should sink type mismatch violation be blocked by content security policy?=] algorithm,
+            passing |global|, |sink|, `"script"`, and the empty string as |source|.
+        1. If |disposition| is `“Allowed”`, return |input|.
+        1. Throw a {{TypeError}}.
+    1. Return |input|.
+1. Let |runScriptsInput| be |input|["runScripts"], [=with default=] false.
+1. Let |argument| be the [=ordered map=] «[ "runScripts" → |runScriptsInput| ]».
+1. Let |sanitizerInputConfig| be the results of [=get a sanitizer config from dictionary|getting a sanitizer config=] from |input|.
+1. If |sanitizerInputConfig| is not null:
+    1. Let |sanitizerInput| be a new {{Sanitizer}}, [=configure|configured=] with |sanitizerInputConfig|.
```
zcorpan: Missing the argument `allowCommentsPIsAndDataAttributes`

spec/index.bs
```diff
@@ -1152,14 +1344,76 @@ To <dfn export>get trusted type compliant string</dfn> a {{TrustedType}} type (|
...
+    whose [=TrustedParserOptions/sanitizer config=] is set to the result of
+    [=get a sanitizer config from dictionary|getting a sanitizer configuration=] from
+    |policyValue|,
+    and whose [=TrustedParserOptions/runScripts=] is set to |runScripts|.
+
+To <dfn>get a sanitizer config from dictionary</dfn> given a dictionary |dictionary|:
+1. Let |sanitizerInput| be |dictionary|["sanitizer"], [=with default|defaulting to=] null.
+1. If |sanitizerInput| is a {{Sanitizer}}, return a [=map/clone=] of |sanitizerInput|'s [=Sanitizer/configuration=].
+1. If |sanitizerInput| is a {{SanitizerPresets}} or a {{SanitizerConfig}}:
+    1. Let |sanitizer| be a new {{Sanitizer}}, [=configure|configured=] with |sanitizerInput|.
```
zcorpan: Missing the argument `allowCommentsPIsAndDataAttributes`

spec/index.bs
```diff
@@ -1134,6 +1311,21 @@ a string |value|, a list |arguments|, and a boolean |throwIfMissing|, perform th
     |args| and `"rethrow"`.
 1.  Return |policyValue|.
 
+## Get Trusted Type compliant input ## {#get-trusted-type-compliant-input-algorithm}
+
+To <dfn export>get trusted type compliant input</dfn> given a [=realm/global object=] |global|, a string or {{TrustedHTML}} |input|,
+a {{TrustedParserOptions}}, {{ParseHTMLUnsafeOptions}} or {{SetHTMLUnsafeOptions}} |options|, and a string |sink|:
+
+1. Let |compliantOptions| be the result of calling [=get trusted type compliant parser options=], passing |global|, |options|, false, and |sink|.
```
zcorpan, line 1317: **`TrustedHTML` input still picks up the policy's sanitizer.** "Get trusted type compliant input"
calls "get trusted type compliant parser options" before checking the input type, so for a
`TrustedHTML` value `compliantOptions` carries the policy's sanitizer config. In an XML document, HTML
then reaches:

> If `mode` is Legacy and `target`'s node document is an XML document, then: If `sanitizer` is
> non-null, then throw a `TypeError`.

...and so once a default policy's `createParserOptions` returns a sanitizer,
`element.innerHTML = trustedHTML` throws unconditionally in XHTML, and `TrustedHTML` values are
sanitized in HTML documents. Returning a `TrustedParserOptions` with a null sanitizer config from the
`TrustedHTML` branch would keep the `runScripts` control without either effect.

spec/index.bs
```diff
@@ -650,16 +692,49 @@ access to which can be controlled using JavaScript techniques
...
 </div>
 
+Note: When enforcing Trusted Types on HTML injection sinks that parse markup (such as {{Element/setHTMLUnsafe()}} or {{Element/innerHTML}}),
+{{TrustedTypePolicyOptions/createParserOptions}} returning a non-null sanitizer configuration allows raw strings to be parsed and sanitized directly,
+without requiring {{TrustedTypePolicyOptions/createHTML}} or a {{TrustedHTML}} instance. For string-only injection sinks that do not support parser options
+(such as {{HTMLIFrameElement/srcdoc}} or `document.write()`), raw strings must still be converted via {{TrustedTypePolicyOptions/createHTML}}.
+
+Note: The `sanitizer` option is only supported for HTML documents. If `createParserOptions`
+returns a dictionary with a non-null `sanitizer` value for an XML document, the sink throws
+an exception. When protecting XML documents, authors are advised to use `createHTML` instead.
```
zcorpan: This note is broader than the behavior; HTML only throws for the Legacy sinks. `setHTMLUnsafe()`, `setHTML()` and `parseHTMLUnsafe()` in an XML document use the HTML parser and do sanitize.

spec/index.bs
```diff
@@ -234,6 +234,13 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
     type: dfn; text: set the inner text steps; url: set-the-inner-text-steps
     type: dfn; text: src; url: attr-script-src
     type: dfn; text: event handler content attribute; url: event-handler-content-attributes
+    type: dfn; text: configuration; for: Sanitizer; url: configuration
+    type: dfn; text: canonicalize the configuration; url: canonicalize-the-configuration
+    type:dictionary; text:Sanitizer
```
zcorpan: should be `type:interface`

spec/index.bs
```diff
@@ -234,6 +234,13 @@ spec: HTML; urlPrefix: https://html.spec.whatwg.org/
...
+    type:dictionary; text:SanitizerConfig
+    type:dictionary; text:SanitizerPresets
```
zcorpan: should be `type:enum`

spec/index.bs
```diff
@@ -1152,14 +1344,76 @@ To <dfn export>get trusted type compliant string</dfn> a {{TrustedType}} type (|
...
+1. If the algorithm threw an error, rethrow the error and abort the following steps.
+1. If |policyValue| is `null` or `undefined`:
+    1. Let |disposition| be the result of executing [=should sink type mismatch violation be blocked by content security policy?=] algorithm,
+        passing |global|, |sink|, `"script"`, and the empty string as |source|.
+    1. If |disposition| is `“Allowed”`, return |input|.
+
+        Note: This step assures that the default policy rejection will be reported, but ignored in a report-only mode.
+
+    1. Throw a {{TypeError}}.
+1. Let |runScripts| be true if |policyValue|["runScripts"] is true and |input| is a {{SetHTMLUnsafeOptions}} whose {{SetHTMLUnsafeOptions/runScripts}} is true.
```
zcorpan: Can't type-check dictionaries.

zcorpan (changes requested):
Opus+Codex review of new changes

spec/index.bs
zcorpan: (Maybe this was intentional?)

spec/index.bs
noamr: Fixed

spec/index.bs
noamr: Fixed

spec/index.bs
noamr: Fixed

spec/index.bs
noamr: Fixed. When it's `TrustedHTML`, we consider the input trusted and don't apply the default sanitizer on it.

zcorpan (approved):

spec/index.bs
```diff
@@ -1152,14 +1346,76 @@ To <dfn export>get trusted type compliant string</dfn> a {{TrustedType}} type (|
...
+    whose [=TrustedParserOptions/sanitizer config=] is set to the result of
+    [=get a sanitizer config from dictionary|getting a sanitizer configuration=] from
+    |policyValue|,
+    and whose [=TrustedParserOptions/runScripts=] is set to |runScripts|.
+
+To <dfn>get a sanitizer config from dictionary</dfn> given a dictionary |dictionary|:
+1. Let |sanitizerInput| be |dictionary|["sanitizer"], [=with default|defaulting to=] null.
+1. If |sanitizerInput| is a {{Sanitizer}}, return a [=map/clone=] of |sanitizerInput|'s [=Sanitizer/configuration=].
+1. If |sanitizerInput| is a {{SanitizerPresets}} or a {{SanitizerConfig}}:
+    1. Let |sanitizer| be a new {{Sanitizer}}, [=configure|configured=] with |sanitizerInput| and true.
```
evilpie: This only handles SanitzerConfig dictionaries correctly and doesn't handle `SanitizerPresets` ("default").

spec/index.bs
```diff
@@ -1152,14 +1346,76 @@ To <dfn export>get trusted type compliant string</dfn> a {{TrustedType}} type (|
...
+    1. Return a [=map/clone=] of |sanitizer|'s [=Sanitizer/configuration=].
+1. Return null.
```
evilpie: Maybe it would be better to default to `{}` ? I think that would allow you to remove this `|| {}` from the example:
```
      const sanitizer = new Sanitizer(options.sanitizer || {});
```

evilpie: Okay, maybe it's not a great idea to default to allowing everything.

noamr: No because we need to retain null for the XML case (throwing).

spec/index.bs
noamr: Fixed

noamr: Fixed

spec/index.bs
```diff
@@ -1068,7 +1068,7 @@ dictionary TrustedTypePolicyOptions {
 callback CreateHTMLCallback = DOMString? (DOMString input, any... arguments);
 callback CreateScriptCallback = DOMString? (DOMString input, any... arguments);
 callback CreateScriptURLCallback = USVString? (DOMString input, any... arguments);
-callback CreateParserOptionsCallback = SetHTMLUnsafeOptions? (optional SetHTMLUnsafeOptions options, any... arguments);
+callback CreateParserOptionsCallback = SetHTMLUnsafeOptions? (SetHTMLUnsafeOptions options, any... arguments);
```
zcorpan, line 1069: Was it intentional to drop `optional`?

noamr: Yes, it came up in one of the reviews.

zcorpan (approved):

spec/index.bs
```diff
@@ -563,6 +573,39 @@ The value is set when the object is created, and will never change during its li
...
+
+The {{TrustedParserOptions}} interface represents options that a developer
+can confidently pass into an [=injection sink=] that will parse HTML.
+These objects are immutable wrappers around a
+{{SetHTMLUnsafeOptions}} dictionary, constructed via a {{TrustedTypePolicy}}'s
+{{TrustedTypePolicy/createParserOptions(options, ...arguments)|createParserOptions}} method.
+
+<pre class="idl">
+[Exposed=Window]
+interface TrustedParserOptions {
```
annevk: Did we discuss whether this should be `TrustedHTMLParserOptions`?

spec/index.bs
```diff
@@ -563,6 +573,39 @@ The value is set when the object is created, and will never change during its li
...
+  [NewObject] Sanitizer? sanitizer();
+  readonly attribute boolean runScripts;
+};
+</pre>
+
+{{TrustedParserOptions}} objects have an associated boolean <dfn export for="TrustedParserOptions">runScripts</dfn>
+and an associated {{SanitizerConfig}}-or-null <dfn export for="TrustedParserOptions">sanitizer config</dfn>.
```
annevk: config -> configuration

spec/index.bs
```diff
@@ -563,6 +573,39 @@ The value is set when the object is created, and will never change during its li
...
+can confidently pass into an [=injection sink=] that will parse HTML.
+These objects are immutable wrappers around a
+{{SetHTMLUnsafeOptions}} dictionary, constructed via a {{TrustedTypePolicy}}'s
+{{TrustedTypePolicy/createParserOptions(options, ...arguments)|createParserOptions}} method.
+
+<pre class="idl">
+[Exposed=Window]
+interface TrustedParserOptions {
+  [NewObject] Sanitizer? sanitizer();
+  readonly attribute boolean runScripts;
```
annevk: I don't really understand why we have this method and getter. We don't expose the details of other trusted objects. Why would we do so here?

spec/index.bs
```diff
@@ -650,16 +693,48 @@ access to which can be controlled using JavaScript techniques
...
+</xmp>
+</div>
+
+Note: When enforcing Trusted Types on HTML injection sinks that parse markup (such as {{Element/setHTMLUnsafe()}} or {{Element/innerHTML}}),
+{{TrustedTypePolicyOptions/createParserOptions}} returning a non-null sanitizer configuration allows raw strings to be parsed and sanitized directly,
+without requiring {{TrustedTypePolicyOptions/createHTML}} or a {{TrustedHTML}} instance. For string-only injection sinks that do not support parser options
+(such as {{HTMLIFrameElement/srcdoc}} or `document.write()`), raw strings must still be converted via {{TrustedTypePolicyOptions/createHTML}}.
+
+Note: When used in legacy HTML injection sinks such as {{Element/innerHTML}}, the `sanitizer` option is only supported for HTML documents.
+When protecting XML documents using legacy sinks, authors are advised to use `createHTML` instead.
```
annevk: Adding this sentence seems a bit dubious given the open issue on removing support for that.

spec/index.bs
```diff
@@ -956,6 +1032,21 @@ Each TrustedTypePolicy object has an associated {{TrustedTypePolicyOptions}} <df
       <dd>arguments</dd>
     </dl>
 
+: <dfn>createParserOptions(options, ...arguments)</dfn>
+::  Returns the
+    result of executing the [=Create a Trusted Type=] algorithm, with the
```
annevk: Should this be lowercase?

spec/index.bs
```diff
@@ -1134,6 +1309,26 @@ a string |value|, a list |arguments|, and a boolean |throwIfMissing|, perform th
     |args| and `"rethrow"`.
 1.  Return |policyValue|.
 
+## Get Trusted Type compliant input ## {#get-trusted-type-compliant-input-algorithm}
+
+To <dfn export>get trusted type compliant input</dfn> given a [=realm/global object=] |global|, a string or {{TrustedHTML}} |input|,
+a {{TrustedParserOptions}}, {{ParseHTMLUnsafeOptions}} or {{SetHTMLUnsafeOptions}} |options|, a string |sink|, and an optional `"html"` or `"xml"` |documentType| (default `"html"`):
+
+1. Let |compliantOptions| be the result of calling [=get trusted type compliant parser options=], passing |global|, |options|, false, and |sink|.
+1. If the algorithm threw an error, rethrow the error.
```
annevk: This is automatic for specifications following Infra.

spec/index.bs
```diff
@@ -1134,6 +1309,26 @@ a string |value|, a list |arguments|, and a boolean |throwIfMissing|, perform th
...
+1. If |input| is an instance of {{TrustedHTML}}:
+    1. Let |runScripts| be |compliantOptions|'s [=TrustedParserOptions/runScripts=] if |compliantOptions| is an instance of {{TrustedParserOptions}}; otherwise |compliantOptions|["runScripts"], [=with default=] false.
+    1. Let |sanitizerConfig| be |options|'s [=TrustedParserOptions/sanitizer config=] if |options| is an instance of {{TrustedParserOptions}}; otherwise the result of [=get a sanitizer config copy from dictionary|getting a sanitizer configuration=] from |global|'s [=relevant realm=] and |options|.
+    1. Let |trustedOptions| be a new {{TrustedParserOptions}} in |global|'s [=relevant realm=] whose [=TrustedParserOptions/sanitizer config=] is |sanitizerConfig| and whose [=TrustedParserOptions/runScripts=] is |runScripts|.
+    1. Return (stringified |input|, |trustedOptions|).
+1. If |documentType| is `"html"`:
```
annevk: The straight quotes typically go on the outside. At least for Infra documents. Might be an existing issue.

spec/index.bs
```diff
@@ -1152,14 +1347,87 @@ To <dfn export>get trusted type compliant string</dfn> a {{TrustedType}} type (|
...
 1. Return stringified |convertedInput|.
 
+## Get Trusted Type compliant parser options ## {#get-trusted-type-compliant-parser-options-algorithm}
+
+Note: |throwIfMissing| is false when called from [=get trusted type compliant input=]. For synchronous fragment injection sinks, a valid sanitizer configuration in the resulting options can satisfy Trusted Types enforcement without requiring {{TrustedTypePolicyOptions/createHTML}} on the input string.
+When streaming or in contexts where the HTML string cannot be validated up front, callers pass true for |throwIfMissing| to ensure a default policy with {{TrustedTypePolicyOptions/createParserOptions}} is required.
+
+To <dfn export>get trusted type compliant parser options</dfn> given a [=realm/global object=] |global|,
+|input|, which is either a {{TrustedParserOptions}}, a {{SetHTMLUnsafeOptions}}, or a {{ParseHTMLUnsafeOptions}},
+a boolean |throwIfMissing|, and a string |sink|, run these steps:
```
annevk:
```suggestion
a boolean |throwIfMissing|, and a string |sink|:
```

annevk (commented):
This largely looks good, but not sure about the name of `TrustedParserOptions` and whether it needs any API.

spec/index.bs
noamr: It was never suggested/contested. I don't feel strongly about it.

spec/index.bs
noamr: The other objects are representing strings and their stringifier exposes that underlying string. This API mirrors that behavior. An alternative would be to turn it into a plain dictionary inside toJSON instead. I am fine either way or with removing the API altogether.

spec/index.bs
noamr: it's like that in the entire spec...

spec/index.bs
noamr: Fixed

spec/index.bs
noamr: Fixed

spec/index.bs
noamr: Removed

spec/index.bs
noamr: Fixed

annevk (approved):

annevk:
When merging this I got confused by the commit message. Is

> When called with `TrustedHTML`, the default policy's `createParserOptions` is used only as a way to disable `runScripts`, and the originally given sanitizer is carried forward.

actually true? I thought we'd just respect whatever sanitizer the policy would come up with? (In general the commit message might need another audit.)

noamr:
> When merging this I got confused by the commit message. Is
> 
> > When called with `TrustedHTML`, the default policy's `createParserOptions` is used only as a way to disable `runScripts`, and the originally given sanitizer is carried forward.
> 
> actually true? I thought we'd just respect whatever sanitizer the policy would come up with? (In general the commit message might need another audit.)



> When merging this I got confused by the commit message. Is
> 
> > When called with `TrustedHTML`, the default policy's `createParserOptions` is used only as a way to disable `runScripts`, and the originally given sanitizer is carried forward.
> 
> actually true? I thought we'd just respect whatever sanitizer the policy would come up with? (In general the commit message might need another audit.)

Revised the commit message to reflect recent changes.

annevk:
> When called with TrustedHTML, the options given as input are carried forward

What does this mean? A naïve reading suggests this means an untrusted sanitizer is respected, which could replace random elements with `script` elements.

noamr:
> > When called with TrustedHTML, the options given as input are carried forward
> 
> What does this mean? A naïve reading suggests this means an untrusted sanitizer is respected, which could replace random elements with `script` elements.

If the caller passes a `TrustedHTML`, it means that they trust it to also run scripts if it's not protected by their given sanitizer.

e.g. it's already downstream-sanitized, or it comes from a trusted source that had direct access to a vetted policy.

annevk:
The caller might be an attacker that got hold of a TrustedHTML value, no? And that value might be safe, but unsafe after replaceWithChildren or some such.

noamr:
> The caller might be an attacker that got hold of a TrustedHTML value, no? And that value might be safe, but unsafe after replaceWithChildren or some such.

OK makes sense. I removed the `TrustedHTML` special case from the spec, WPT and PR description

noamr:
> > The caller might be an attacker that got hold of a TrustedHTML value, no? And that value might be safe, but unsafe after replaceWithChildren or some such.
> 
> OK makes sense. I removed the `TrustedHTML` special case from the spec, WPT and PR description

This is actually an existing issue. See #615.
However removing the TrustedHTML special case is justified given the other comments about preferring TrustedHTMLOptions and using createHTML and TrustedHTML only as a fallback.

Also this offers authors some mitigation for #615 because the default policy still applies for TrustedHTML.

